Issue #82, opened 2024-08-08
Docs: Add instructions on how to specify and use secrets


kimdre:
Secrets should never be committed to your Git Repositories.

Instead either
- use secret files and mount them to a specifc destination inside the doco-cd container  
   https://docs.docker.com/compose/use-secrets/
- or use docker secrets and let docker store the secrets for you.  
   https://docs.docker.com/engine/swarm/secrets/

maxexcloo:
I think it would be cool to be able to assign secrets using the tool - I currently push docker secrets to the server using terraform (they’re generated there) and it would be nice to read them out using this tool to avoid having to mess around with the images themselves - not sure how this would work though, what was your idea for the second option?

kimdre:
Currently you can use these examples to use secrets:
- https://docs.docker.com/compose/use-secrets/
- https://stackoverflow.com/a/42151570

PR #85, opened 2024-08-09
feat: add support for data encryption via SOPS

kimdre:
This PR adds support for using encrypted data in deployments via [SOPS](https://github.com/getsops/sops).

SOPS is designed to encrypt structured data (like yaml, json, env) so that it can be shared or uploaded to a Git Repository without leaking sensitive information.

Related to #82


kimdre:
Currently blocked because of weird errors when using the SOPS package...
https://github.com/getsops/sops/issues/1582

kimdre:
## To-Dos:

- [ ] Finish logic for unencrypting secret for deployments
- [ ] Create new GPG Keypair and add it to Github Action secrets
- [ ] Add new testfiles that are encrypted with the new GPG key
- [ ] Adjust test workflow to use GPG Keypair

Issue #275, opened 2025-02-10
Can't mount repository files in volumes


isc30:
Hi, I'm trying to deploy a test nginx page using doco-cd but it seems like it doesn't pull all the repository files to mount them as volumes.

## what works:

docker-compose.yml
```yaml
services:
  helloworld:
    image: nginx
    ports:
      - 8080:80
```

This works fine, and the nginx "Welcome Page" is rendered on `http://dockerhost:8080`.

![Image](https://github.com/user-attachments/assets/c3a17316-33b2-4951-ace8-768588fa3e27)

## what doesn't work:

If I add a new `index.html` file in the repo and try to serve it, it fails with 403:

html/index.html
```html
Hello World!
```

docker-compose.yml
```yaml
services:
  helloworld:
    image: nginx
    ports:
      - 8080:80
    volumes:
      - ./html:/usr/share/nginx/html
```

![Image](https://github.com/user-attachments/assets/1af0f891-a478-450e-abc0-2d4469fdb2e1)

Going inside the helloworld container and executing `ls /usr/share/nginx/html` shows that there are no files mounted there.

It seems the bug could be caused by:
- maybe `doco-cd` is not pulling the file from the repository correctly
- maybe after running `docker compose up`, the repository that is cloned in a temporary folder is deleted and the files stop being available to the container
- maybe the file has incorrect permissions (im using `group_add` in the doco-cd config) and the nginx container can't read the file

kimdre:
I think your issue is related to #172

isc30:
Hi @kimdre it doesn't seem like its the same thing, in my case the example is way simpler and I'm not mounting any file from one container to the other. I'm just mounting an `index.html` that exists in the repo, but because the `git clone` happens in a temporary folder, the file isn't available.

kimdre:
Hi, sorry for being away for quite a while. I think I now fixed the problem. The issue was a flaw in how I implemented the deployment logic, because Docker Engine is not able to reach the cloned files inside the container. The app now uses a volume or bind mount as the deployment directory, so that Docker Engine can reach them from outside. Because of this, the fix contains breaking changes that require some adjustments inside your docker-compose.yml.

- I published a new [pre-release](https://github.com/kimdre/doco-cd/releases/tag/v0.13.0-rc.1) version [v0.13.0-rc.1](https://github.com/kimdre/doco-cd/tree/refs/tags/v0.13.0-rc.1) that you can use to test if everything is working now for you. 
[...]

Issue #172, opened 2024-10-11
doco-cd tries to run test.compose.yml after triggering webhook


Staubgeborener:
First of all, this is my `docker-compose.yml`:
```
user@server::~/docker/gitops$ cat docker-compose.yml
services:
  app:
    container_name: doco-cd
    image: ghcr.io/kimdre/doco-cd:latest
    restart: unless-stopped
    ports:
      - "82:80"
    environment:
      GIT_ACCESS_TOKEN: my_github_token
      WEBHOOK_SECRET: my_webhook_secret
    group_add:
      - "996"
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
```
Behind port 82 is a reverse proxy which is listening for webhooks on a subdomain. Lets call it `sub.domain.com`

This is my docker log for doco-cd:
```
{"time":"2024-10-11T19:06:48.19927499Z","level":"debug","msg":"repository cloned","job_id":"1b53aced-0fc7-4fae-89ef-29e97d6a680b","repository":"Staubgeborener/gitops","path":"/tmp/Staubgeborener/gitops"}
{"time":"2024-10-11T19:06:48.199332994Z","level":"debug","msg":"retrieving deployment configuration","job_id":"1b53aced-0fc7-4fae-89ef-29e97d6a680b","repository":"Staubgeborener/gitops"}
{"time":"2024-10-11T19:06:48.199745173Z","level":"debug","msg":"deployment configuration retrieved","job_id":"1b53aced-0fc7-4fae-89ef-29e97d6a680b","repository":"Staubgeborener/gitops","stack":"test","reference":"refs/heads/main","config":{"Name":"test","Reference":"refs/heads/main","WorkingDirectory":".","ComposeFiles":["test.compose.yml"],"RemoveOrphans":true,"ForceRecreate":false,"ForceImagePull":false,"Timeout":180,"BuildOpts":{"ForceImagePull":false,"Quiet":false,"Args":null,"NoCache":false}}}
{"time":"2024-10-11T19:06:48.211445635Z","level":"info","msg":"deploying stack","job_id":"1b53aced-0fc7-4fae-89ef-29e97d6a680b","repository":"Staubgeborener/gitops","stack":"test","reference":"refs/heads/main"}
[...]

PR #290, opened 2025-02-25
fix(#275): fix volume mounts and binds

kimdre:
This PR fixes volume mounts and binds in deployments using docker hooks.

The problem was, that the tmp directory was cleaned up at the end of the deployment, even though the container wanted to access these files (See #275 and #172)

Now the tmp directory cleaned up on container events that take every scenario into account.


kimdre:
I just tested it a little and so far volumes are working even after removing/restarting the doco-cd container or restarting my test nginx container, but I need to remove the doco-cd container and do a fresh start after every deployment because the cloned repo already exists. 
Maybe it's probably enough if in that case doco-cd just does a git checkout to the commit sha in the payload? 

What does not seem to work is binding files from the deployed repo to the container, like a index.html that I want to mount to the nginx container.

So something like this does not work yet for some reason:
```yaml
volumes:
     - ./index.html:/usr/share/nginx/html/index.html
```

But writing the file in a volume works.

kimdre:
Bind mounts are currently still not working. I need to implement a solution that uses a mount from the host filesystem to the doco-cd container (bind or volume) so that the files in the cloned repo can be reached.
https://github.com/docker/compose/issues/12801

kimdre:
The container user must be root now to be able to access its data directory (The Docker Host binds container volumes and mounts from outside the doco-cd container).

## TODO
- Change Dockerfile base image from nonroot to root
- Remove group_add from docker-compose files (dev and examples)
- Remove nonroot example in wiki

kimdre:
### Breaking changes with this PR
- The container now requires volume or bind mount for its data.
- non-root user has been removed. The container user must always be root now to be able to access its data directory (The Docker Host binds container volumes and mounts from outside the doco-cd container).

PR #490, opened 2025-07-08
feat(docker-swarm): add config and secret hash prefixing to overwrite deployed configs and secrets

kimdre:
To Do:

- [x] Append a content hash of configs/secrets to their names as a suffix to avoid name collisions when deploying new version of them with changed content (docker swarm only) with new deploy var `disable_name_suffix_hash` (bool) to disable this behavior if needed. 
- [ ] Add `disable_name_suffix_hash` deploy setting to wiki
- [ ] Deploy with ForceRecreate enabled when the Git status contains a file that is referenced as the source of a config or secret.
- [ ] Deploy with ForceRecreate enabled when the inline content of a config changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ test-coverage:
 # Run specified tests from arguments
 test-run:
 	@echo "Running tests: $(filter-out $@,$(MAKECMDGOALS))"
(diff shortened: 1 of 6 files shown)


kimdre:
Closed because doco-cd currently does not fully support swarm-mode.

PR #546, opened 2025-08-09
fix: trim surrounding whitespaces from secret file contents

kimdre:


nbx3:
Just tried this and still an issue, had a look at code, wondering if the line shouldn't be 

```
*field.value = strings.TrimSpace(*field.value)
```

I dont do much golang, but use C from time to time so i think pointer knowledge transfers!

kimdre:
@nbx3 Sorry, I can't replicate it. It works for me, even when I add multiple newlines before and after the secret in the file.
Can you show me how your setup looks like?

nbx3:
That is weird, i'll get all the versions i have in a few minutes, just have to run some errands, but what i was also trying to point out was that the current trim function is triming the filename field (`field.fileField`) not the value field (`field.value`) unless they have the same values?

kimdre:
@nbx3 The naming is just a little weird, because the env module actually writes the file content to the `*File` variables when parsed, not the name of the env name so it's actually correct. :D

nbx3:
Try this

```
package main

import (
    "fmt"
    "strings"
    "github.com/caarlos0/env/v11"
)

const ErrBothSecretsSet = "Both secrets are set"
const ErrBothSecretsNotSet = "Both secrets not set"

type AppConfig struct {
    GitAccessToken      string                 `env:"GIT_ACCESS_TOKEN"`
    GitAccessTokenFile  string                 `env:"GIT_ACCESS_TOKEN_FILE,file"`
    WebhookSecret       string                 `env:"WEBHOOK_SECRET"`
    WebhookSecretFile   string                 `env:"WEBHOOK_SECRET_FILE,file"`
    LogLevel            string                 `env:"LOG_LEVEL,notEmpty" envDefault:"info"`
}

// loadFileBasedEnvVars loads environment variables from files if the corresponding file-based environment variable is set.
func loadFileBasedEnvVars(cfg *AppConfig) error {
  fields := []struct {
    fileField string
    value     *string
    name      string
  }{
    {cfg.WebhookSecretFile, &cfg.WebhookSecret, "WEBHOOK_SECRET"},
    {cfg.GitAccessTokenFile, &cfg.GitAccessToken, "GIT_ACCESS_TOKEN"},
  }

  for _, field := range fields {
    if field.fileField != "" {
      if *field.value != "" {
        return fmt.Errorf("%w: %s or %s", ErrBothSecretsSet, field.name, field.name+"_FILE")
      }

      *field.value = strings.TrimSpace(field.fileField)
    } else if *field.value == "" {
      return fmt.Errorf("%w: %s or %s", ErrBothSecretsNotSet, field.name, field.name+"_FILE")
    }
  }

  return nil
}


func main() {

    cfg := AppConfig{}

    if err := env.Parse(&cfg); err != nil {
        fmt.Printf("Env parse: %s\n", err)
    }

    if err := loadFileBasedEnvVars(&cfg); err != nil {
        fmt.Printf("File parse: %s\n", err)
    }

    fmt.Printf("GitAccessToken: %s\n", cfg.GitAccessToken)
    fmt.Printf("GitAccessTokenFile: %s\n", cfg.GitAccessToken)
    fmt.Printf("WebhookSecret: %s\n", cfg.WebhookSecret)
    fmt.Printf("WebhookSecretFile: %s\n", cfg.WebhookSecretFile)
    fmt.Printf("LogLevel: %s\n", cfg.LogLevel)

}
```

```
echo 123 > webhook_secret
echo 321 > git_access_token
```

These are my results - 

```
GIT_ACCESS_TOKEN_FILE=git_access_token WEBHOOK_SECRET_FILE=webhook_secret go run .
GitAccessToken: 321
GitAccessTokenFile: 321
WebhookSecret: 123
WebhookSecretFile: 123

LogLevel: info
```

An extra newline somehow is being inserted at the end of the webooksecretfile, but i can't see where, can you?

kimdre:
@nbx3 Can you please try to write your secrets without newlines? Either `echo -n 123 > webhook_secret` or `printf 123 > webhook_secret`.

The values of the *File variables are only used to write the file content to the other non-File variable and is not used anymore after that, so the values in your example are all correct. You can see how the `WebhookSecret` does not contain the newline character of `WebhookSecretFile`.

nbx3:
Thanks, yea i see what your saying, that does give the correct result from the above. However i'm still intrigued why it only shows the extra newline webhook secret file not the github token one, but the main issue appears to be resolved in that the actual secret value is being set without the newline. 

Im on v0.31.2 but i'm still getting the same issue with "HMAC verification failed" so there must be something in my config somewhere or elsewhere in the code thats causing this. But i'll leave this pull discussion now and if i find something new will open it. Thanks for your help on this.

PR #650, opened 2025-09-30
feat: enable docker healthcheck

cduchenoy:
We can also add in wiki if user change HTTP_PORT to add in docker-compose.yml

```yaml
---
services:

  doco-cd:
    image: ghcr.io/kimdre/doco-cd:0.39.0-rc.1
...
    environment:
      HTTP_PORT: 8080
...
    healthcheck:
      test: ["CMD", "wget", "--no-verbose", "--tries=1", "--spider", "http://localhost:8080/v1/health"]
      start_period: 15s
      interval: 30s
      timeout: 5s
      retries: 3
```


kimdre:
Thank you for the PR!. :) I added the http port to the healthcheck so people know what port is used by default.

PR #729, opened 2025-11-02
feat(poll): add run_once flag to deploy only once via poll job

p0lycarpio:
# The need

This pull request addresses a specific need: initializing stacks once when doco-cd is launched, for example, when setting up a new infrastructure that will then be updated via webhook. This allows for an initial deployment without having to send a test webhook.

# Changes

This pull request introduces an improvement to the polling handler logic to better handle negative interval values, and updates the documentation submodule. The most important changes are:

Polling behavior enhancements:
* Updated the `PollHandler` function in `handler_poll.go` to treat negative `interval` values as a "run once" signal, causing the handler to exit after a single poll. This prevents repeated polling when a negative interval is configured, and adds a debug log for clarity.
* Ensured that the handler only sleeps when a positive interval is set, clarifying behavior for zero or negative intervals.

Documentation:
* Updated the `docs` submodule to the latest commit, reflecting the most recent documentation changes.


kimdre:
Hi, thank you for your PR! I like the idea, but using negative intervals for this feels not right to me. Can you instead add a new boolean `run_once` field to the Poll Config that does what you thought of? :)

p0lycarpio:
Hi, sure! I made the change to include `run_once` in the configuration file. I left [the sleep condition based on positive value](https://github.com/kimdre/doco-cd/pull/729/files#diff-5e072a967f70faa328aa94619b6bd98091f65a240fc625ca9fb4f7492bb55ae1R126)  to avoid looping if a negative value is set.

kimdre:
I removed the sleep condition. The interval value gets already verified beforehand so this is not needed.

kimdre:
Thanks again for your PR!

PR #737, opened 2025-11-07
fix(poll): run_once if interval is set to 0

p0lycarpio:
This pull request makes a minor adjustment to the polling logic in `handler_poll.go`. Now, the polling job will only be disabled if the interval is zero and the `RunOnce` flag is not set, allowing for initial polling jobs even when the interval is zero.

- Polling Logic Update:
  * Modified the condition in `StartPoll` to ensure that polling jobs with `RunOnce` enabled are not disabled when the interval is zero. (`cmd/doco-cd/handler_poll.go`)


kimdre:
Thanks for the PR. Looks good to me. :)

PR #747, opened 2025-11-08
feat(config): add support for inline deploy config in poll configs

### LINKED ISSUE #740: Is the doco-cd.yml file required if using poll configuration?
### Description

I'd like to avoid sprinkling doco-cd.yml configurations across every repo if I'm going to configure every repo explicitly in the doco-cd docker-compose.yml. If I stick to defaults, will it work or is the yml file strictly necessary? Thoughts on enabling full configuration through poll configuration and avoiding doco-cd.yml?

### Use Case

Can automate deployment workflow by only introducing doco-cd without having to introduce a very small yml file across dozens of repos.

### Additional Information

_No response_

tldev:
I think this is a more acceptable v1 - looking for feedback.

closes #740


kimdre:
Looks good to me so far. Let me do some tests. :)

kimdre:
Thank you for the PR!

PR #794, opened 2025-11-20
feat: Allow specifying DEPLOY_CONFIG_BASE_DIR

bjw-s:
The goal of this PR is to make it so that users can specify where doco-cd looks for its deployment configurations through the `DEPLOY_CONFIG_BASE_DIR` environment variable.

**Rationale:**
I have a monorepo (https://github.com/bjw-s-labs/home-ops/) that I use to configure both my Kubernetes homelab _and_ my TrueNAS configuration. I would like to make it so that I can tell doco-cd to fetch its deployment configuration from the `docker/truenas` subfolder of that repo instead of having to create the deployment configuration files in the repo root.


kimdre:
Thanks for the PR. Your current implementation means that even when `DEPLOY_CONFIG_BASE_DIR` is set, you still need to set the `working_dir` relative to the repository root, which might actually be useful if you want to store the deploy configs in a separate directory.

bjw-s:
> Your current implementation means that even when `DEPLOY_CONFIG_BASE_DIR` is set, you still need to set the `working_dir` relative to the repository root, which might actually be useful if you want to store the deploy configs in a separate directory.

I wish I could say that was intentional, but I just completely forgot about that aspect 🫣 
Now that you point it out I can see value in it as well, but I guess that's mostly a matter of updating the documentation wiki to explain the behavior?

kimdre:
The feature is released with https://github.com/kimdre/doco-cd/releases/tag/v0.51.0

Issue #799, opened 2025-11-21
Fail to parse symlinks when running decrypt routine


trevorswanson:
### Description

I have a configuration like the following:

```
.doco-cd.yml
apps/
├── my-app-1/
│   ├── templates@ -> ../../common/templates
│   ├── docker-compose.yml
│   └── .env
common/
├── templates/
│   ├── web-app.yml
```

There is a symlink located at `apps/my-app-1/templates` that points to `common/templates`, which lets me leverage the `include` or `extend` directives in `docker-compose` more easily.

My `common/templates/web-app.yml` looks like this:

```
services:
  web-app:
    networks:
      - traefik
    expose:
      - ${APP_PORT}
    labels:
      traefik.enable: true
      traefik.http.routers.${APP_NAME}.rule: Host(`${APP_NAME}.${APP_BASEDOMAIN}`)
```

And `apps/my-app-1/docker-compose.yml` looks like this:

```
services:
  my-app-1:
    extends:
      file: templates/web-app.yml
      service: web-app
    image: strm/helloworld-http
    container_name: hello-world
```

`apps/my-app-1/.env` has:

```
APP_NAME=hello-world
APP_BASEDOMAIN=lab.local
APP_PORT=80
```

My `.doco-cd.yml` deploy config is in the "deploy config" section of the PR template

Unfortunately when doco-cd attempts to scan for encrypted secrets, I end up with an error listed (included further down in the PR template).

### Steps to reproduce the issue

```bash
1. Create a symlink from one directory to another
2. Attempt to deploy stack with doco-cd
3. See error
```

### Operating System

Linux

### Docker Compose File

```yaml
x-poll-config: &poll-config
  POLL_CONFIG: |
    - url: https://git.lab.local/tswan/dococd-apps.git

services:
  doco-cd:
    container_name: doco-cd
    image: ghcr.io/kimdre/doco-cd:latest
    restart: unless-stopped
    environment:
      TZ: America/New_York
      GIT_ACCESS_TOKEN: XXX
      <<: *poll-config
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
      - ./data:/data
```

### Deployment Config

```yaml
name: hello-world
working_dir: apps/my-app-1
env_files:
  - .env
```

### Logs

```sh
[...]

PR #800, opened 2025-11-21
fix: don't attempt to decrypt symlinks

trevorswanson:
Resolves https://github.com/kimdre/doco-cd/issues/799

When doco-cd is walking the working directory to look for secrets, it correctly [skips directories](https://github.com/kimdre/doco-cd/blob/db2f20ca38f194cf4580f5f6dcdf00e8a7a59427/internal/encryption/decrypt.go#L74-L76), as a directory cannot be encrypted/decrypted. However, it does not currently skip symlinks, which also cannot be encrypted/decrypted.

This PR adds logic to look for the ModeSymlink flag on files, so that we can properly skip them.

**Important note on behavior here:** `filepath.WalkDir` **does not follow symlinks.** This means that, while this PR allows the symlink to exist rather than erroring out as seen in the linked issue, it will not decrypt any secrets found on the other side of the symlink.

I believe this is still an improvement over the current behavior but we may want to add a note in the documents explaining this behavior.

### DIFF
--- a/internal/encryption/decrypt.go
+++ b/internal/encryption/decrypt.go
@@ -75,6 +75,11 @@ func DecryptFilesInDirectory(dirPath string) ([]string, error) {
 			return nil
 		}
 
+		fileInfo, err := os.Lstat(path)
+		if fileInfo.Mode()&os.ModeSymlink == os.ModeSymlink {
+				return nil
+		}
+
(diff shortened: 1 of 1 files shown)


trevorswanson:
https://github.com/kimdre/doco-cd/pull/801 is a much more thorough approach, closing my PR as it's no longer needed

PR #802, opened 2025-11-21
feat(poll): allow auto_discover in inline deployment config

bjw-s:
My previous PR somehow did not work entirely as intended. The DEPLOY_CONFIG_BASE_DIR path logic was leading to `working_dir` being duplicated in the path. I have tried to implement a proper fix and add some tests to address this.

Additionally, I've updated the POLL_CONFIG logic to also add support for auto_discover (the working_dir was not properly working there because it was being passed through 1:1).

Feel free to make any modifications that you feel are necessary and thanks for merging my previous PR!


thomaswienecke:
I'd be interested in this. Is there something blocking this?

kimdre:
Hi, sorry I was very busy recently. I will check your PR soon. :)

PR #900, opened 2025-12-19
feat: respect .gitignore when decrypting files

clstb:
Cool project 🙏 

I think doco should respect .gitignore. I had a service container that created a symlink in a volume (which is .gitignored) which broke as it led outside the repository and doco tried to decrypt it.


kimdre:
Hi, thank you for your PR! :) I made some adjustments to the code

PR #907, opened 2025-12-23
feat(git): add git submodules support

mechanicalbot:
### Summary
- Add a GIT_CLONE_SUBMODULES option to clone/update submodules during repository checkout.
- Wire the option through webhook, poll, and init flows so private submodule repos can be fetched.

### Motivation
My deployments need secrets stored in a private submodule; cloning with submodules ensures those files are available at deploy time.


kimdre:
Thank you for your PR! :) I changed the submodule cloning to be enabled by default.

PR #922, opened 2026-01-10
fix: allow PollConfig inline deployments to use auto_discover settings

thomaswienecke:
Implements the autodiscover logic for inline deployment configs when using the PollConfig.

I saw the PR #802 and since it didn't have much traction for a few weeks and had 2 failing tests I thought I'd give it a shot focusing on only getting the auto_discover settings working when using the inline deployment config. Feel free to close this PR if you think the other PR is more promising.

### DIFF
--- a/internal/config/deploy_config.go
+++ b/internal/config/deploy_config.go
@@ -297,7 +297,29 @@ func validateUniqueProjectNames(configs []*DeployConfig) error {
 func ResolveDeployConfigs(poll PollConfig, configDir, name string) ([]*DeployConfig, error) {
 	// Prefer inline deployments when present
 	if len(poll.Deployments) > 0 {
-		return poll.Deployments, nil
+		configs := poll.Deployments
+
+		// Handle autodiscover for inline deployment configs
+		for i, c := range configs {
+			// Check for deployConfigs with AutoDiscover enabled, if true then remove this config and add new configs based on discovered compose files
+			if c.AutoDiscover {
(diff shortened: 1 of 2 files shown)


bjw-s:
I don't mind the alternative implementation suggestion as I assume your solution will work perfectly fine as well. But I did want to comment on two details in your PR summary:

> didn't have much traction for a few weeks 

I don't think that is really due to my PR, but more likely due to the maintainer not having had the ability to get to it yet.

> and had 2 failing tests

The failing tests are unrelated to the changes. The tests are also failing in `main` with an error of `missing access token for private repository`. You will more than likely see the same errors once the workflows run against your changes.

thomaswienecke:
> > didn't have much traction for a few weeks
> 
> I don't think that is really due to my PR, but more likely due to the maintainer not having had the ability to get to it yet.
> 
Hey, yeah definitely. And I never meant it in a blaming or complaining way. I just wanted to give an option that is a smaller increment hoping the maintainer might have the capacity to review it quicker. However, I'm more than happy if your change makes it because it contributes more than this one.

> > and had 2 failing tests
> 
> The failing tests are unrelated to the changes. The tests are also failing in `main` with an error of `missing access token for private repository`. You will more than likely see the same errors once the workflows run against your changes.

Good point. The logs didn't load for me first (probably due to github loading them from an archive), therefore I just started this PR and only afterwards noticed these pipeline failures were not related to your changes :D

PR #953, opened 2026-01-31
fix(stages): use actual job trigger instead of hardcoded "poll" value

danroc:
The `triggerEvent` parameter was hardcoded to "poll" in `RunDeployStage`, causing webhook-triggered deployments to be treated as poll events. This resulted in unnecessary forced recreates when compose files changed.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,10 +1,5 @@
 name: Build Image
 
-permissions:
-  contents: write # read for actions/checkout to check out the repository and write to upload the SBOM file
-  id-token: write # for docker/login-action to authenticate to the GitHub Container Registry
-  packages: write # for docker/build-push-action to push the image to the GitHub Container Registry
-
 on:
   push:
     tags:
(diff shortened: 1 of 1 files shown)


internal/stages/stage_3_deploy.go
```diff
@@ -66,7 +66,7 @@ func (s *StageManager) RunDeployStage(ctx context.Context, stageLog *slog.Logger
 
 	err = docker.DeployStack(stageLog, s.Repository.PathInternal, s.Repository.PathExternal, &ctx, &s.Docker.Cmd, s.Docker.Client,
 		s.Payload, s.DeployConfig, s.DeployState.ChangedFiles, latestCommit, config.AppVersion,
-		"poll", forceDeploy, s.DeployState.ResolvedSecrets, s.DeployState.SecretsChanged)
+		string(s.JobTrigger), forceDeploy, s.DeployState.ResolvedSecrets, s.DeployState.SecretsChanged)
```
danroc: I made the smallest possible change, but I can also refactor `DeployStack` to accept a `JobTrigger` instead of a string if you prefer.

PR #964, opened 2026-02-05
fix: respect DOCKER_SWARM_FEATURES=false

wajeht:
I'm running a single Docker host in swarm mode with this structure:

```
apps/
  ├── swarm/      # doco-cd with SWARM=true
  ├── compose/    # doco-cd with SWARM=false
  │   ├── plex/       # needs /dev/dri
  │   └── vpn-qbit/   # needs /dev/net/tun
  └── infra/
      ├── doco-cd/          # deploys apps/swarm as swarm stacks
      └── doco-cd-compose/  # deploys apps/compose as compose (SWARM=false)
```

I need two `doco-cd` instances because some apps need device access which swarm doesn't support. But when I set `DOCKER_SWARM_FEATURES=false`, it still deploys as swarm stacks instead of compose.

The issue is that `main.go` sets `swarm.ModeEnabled = true` at startup, and the handlers only re-check when `DockerSwarmFeatures=true`. When it's false, nothing happens and `ModeEnabled` stays true.

This fix adds an else clause to explicitly set s`warm.ModeEnabled = false`.

### DIFF
--- a/.github/workflows/build-dev.yaml
+++ b/.github/workflows/build-dev.yaml
@@ -74,7 +74,7 @@ jobs:
           severity: 'CRITICAL,HIGH'
 
       - name: Upload Trivy scan results to GitHub Security tab
-        uses: github/codeql-action/upload-sarif@6bc82e05fd0ea64601dd4b465378bbcf57de0314 # v4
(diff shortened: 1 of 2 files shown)


kimdre:
HI thanks for your bugfix :) I made some small adjustments.

PR #977, opened 2026-02-08
fix: scope auto-discover cleanup by working directory

wajeht:
When running two doco-cd instances watching different subdirectories of the same repo, the auto-discover cleanup treats the other instance's stacks as obsolete and tries to destroy them every poll cycle.

**My setup:**
```
apps/
├── swarm/    # auto-deployed by doco-cd (SWARM=true, 30+ stacks)
├── compose/  # auto-deployed by doco-cd-compose (SWARM=false, 2 stacks)
└── infra/    # manual deploy, not auto-discovered
    ├── doco-cd/          # watches apps/swarm/
    └── doco-cd-compose/  # watches apps/compose/
```

**doco-cd (swarm):**
```yaml
DOCKER_SWARM_FEATURES: "true"
POLL_CONFIG: |
  - url: https://github.com/wajeht/home-ops.git
    reference: refs/heads/main
    interval: 60
    deployments:
      - working_dir: apps/swarm
        auto_discover: true
        auto_discover_opts:
          delete: true
```

**doco-cd-compose:**
```yaml
DOCKER_SWARM_FEATURES: "false"
POLL_CONFIG: |
  - url: https://github.com/wajeht/home-ops.git
    reference: refs/heads/main
    interval: 60
    deployments:
      - working_dir: apps/compose
        auto_discover: true
        auto_discover_opts:
          delete: true
```

Both instances share the same repo URL. The cleanup queries ALL containers with `cd.doco.deployment.auto_discover=true` and filters only by repo URL, so the compose instance marks every swarm stack as "obsolete" (and vice versa).

**Fix:** Also check `cd.doco.deployment.working_dir` during cleanup. Only consider stacks under the same base path as the current instance's deployments.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1@sha256:b6afd42430b15f2d2a4c5a02b919e98a525b785b1aaff16747d2f623364e39b6
-FROM golang:1.25.7@sha256:011d6e21edbc198b7aeb06d705f17bc1cc219e102c932156ad61db45005c5d31 AS prerequisites
+FROM golang:1.25.7@sha256:cc737435e2742bd6da3b7d575623968683609a3d2e0695f9d85bee84071c08e6 AS prerequisites
 
 ARG APP_VERSION=dev
 ARG DISABLE_BITWARDEN=false


cmd/doco-cd/utils.go
```diff
@@ -69,9 +70,20 @@ func cleanupObsoleteAutoDiscoveredContainers(ctx context.Context, jobLog *slog.L
...
+	// deployed from the same base directory (e.g., "apps/swarm" or "apps/compose").
+	// Without this, multiple doco-cd instances watching different subdirectories
+	// of the same repo would treat each other's stacks as obsolete.
+	workingDirBases := make(map[string]bool)
+
 	for _, cfg := range deployConfigs {
 		if cfg.AutoDiscover {
 			autoDiscoveredNames[cfg.Name] = cfg.AutoDiscoverOpts.Delete
+
+			base := filepath.Dir(cfg.WorkingDirectory)
```
kimdre: filepath.Dir returns the path without the last element, so your `base` variable would be "apps" for both of your working directories (`apps/swarm` and `apps/compose`).

kimdre (changes requested):
HI, thanks for your PR. Please fix your code as it contains a bug.

PR #998, opened 2026-02-14
feat(external-secrets): add generic webhook-based secrets provider

UiP9AV6Y:
this secret provider has been modeled after the external-secrets operator provider of the [same name][1]. the implementation differs in a few ways:

* JMESPath instead of JSONPath for result parsing. the former is designed for querying while the latter has transformation capabilities, which are not needed for simply reading a value from a JSON document.
* no header support. configuring arbitrary key-value pairs for HTTP headers via environment variables is not straight forward which is why it has not been implemented as part of this MR.

[1]: https://external-secrets.io/latest/provider/webhook/

### DIFF
--- a/internal/docker/compose_test.go
+++ b/internal/docker/compose_test.go
@@ -181,7 +181,11 @@ func TestDeployCompose(t *testing.T) {
 		t.Fatalf("Failed to get auth method: %v", err)
 	}
 
-	t.Logf("Using auth method: %s", auth.Name())
+	if auth != nil {
+		t.Logf("Using auth method: %s", auth.Name())
+	} else {
+		t.Log("No auth method configured, using anonymous access")
+	}
 
 	repo, err := git.CloneRepository(tmpDir, p.CloneURL, p.Ref, c.SkipTLSVerification, c.HttpProxy, auth, c.GitCloneSubmodules)
 	if err != nil {
@@ -386,7 +390,11 @@ func TestHasChangedConfigs(t *testing.T) {
 		t.Fatalf("Failed to get auth method: %v", err)
 	}
 
-	t.Logf("Using auth method: %s", auth.Name())
+	if auth != nil {
+		t.Logf("Using auth method: %s", auth.Name())
+	} else {
+		t.Log("No auth method configured, using anonymous access")
+	}
 
 	tmpDir := t.TempDir()
 
@@ -456,7 +464,11 @@ func TestHasChangedSecrets(t *testing.T) {
 		t.Fatalf("Failed to get auth method: %v", err)
 	}
 
-	t.Logf("Using auth method: %s", auth.Name())
+	if auth != nil {
+		t.Logf("Using auth method: %s", auth.Name())
+	} else {
+		t.Log("No auth method configured, using anonymous access")
+	}
 
 	tmpDir := t.TempDir()
 
@@ -526,7 +538,11 @@ func TestHasChangedBindMounts(t *testing.T) {
 		t.Fatalf("Failed to get auth method: %v", err)
 	}
 
-	t.Logf("Using auth method: %s", auth.Name())
+	if auth != nil {
+		t.Logf("Using auth method: %s", auth.Name())
+	} else {
+		t.Log("No auth method configured, using anonymous access")
+	}
 
 	tmpDir := t.TempDir()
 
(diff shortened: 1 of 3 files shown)


UiP9AV6Y:
i can also provide a follow up MR with an example implementation for a webhook provider service akin to the [external-secrets fake](https://external-secrets.io/latest/provider/fake/) provider

kimdre:
Hi thanks for your PR! :) 

Can you please write a comment with documentation how to use the new webhook-based provider similiar to the providers [in the wiki](https://github.com/kimdre/doco-cd/wiki/External-Secrets)?

UiP9AV6Y:
## Webhook

To use fetch secrets via the generic webhook provider, you need to set the following environment variables:

| Key                                   | Value                                                                                                                |
|---------------------------------------|----------------------------------------------------------------------------------------------------------------------|
| `SECRET_PROVIDER`                     | `webhook`                                                                                                            |
| `SECRET_PROVIDER_SITE_URL`            | URL [template][Golang Text Template] to query for secrets (required)                                                 |
| `SECRET_PROVIDER_RESULT_JMES_PATH`    | [JMESPath][] query to run against the webhook response (required)                                                    |
| `SECRET_PROVIDER_REQUEST_BODY`        | HTTP Body JSON [template][Golang Text Template] to send with requests                                                |
| `SECRET_PROVIDER_BEARER_TOKEN`        | Bearer token to send with the request in the Authorization header                                                    |
| `SECRET_PROVIDER_BEARER_TOKEN_FILE`   | Path to the file containing the authentication bearer token                                                          |
| `SECRET_PROVIDER_BASIC_USERNAME`      | Username for the Basic Auth scheme to send with the request in the *Authorization* header                            |
| `SECRET_PROVIDER_BASIC_USERNAME_FILE` | Path to the file containing the authentication username                                                              |
| `SECRET_PROVIDER_BASIC_PASSWORD`      | Password for the Basic Auth scheme to send with the request in the *Authorization* header                            |
| `SECRET_PROVIDER_BASIC_PASSWORD_FILE` | Path to the file containing the authentication password                                                              |

**Deployment configuration**

Add a mapping/reference between the environment variable you want to set in the docker compose project/stack and the ID of the secret the remote webhook provides.

For example in your `.doco-cd.yml`:

```yaml
name: myapp
external_secrets:
  DB_PASSWORD: this-can-be-included-in-the-request-as-remoteRef
```

Both the webhook URL and request body will be rendered as [Golang Text Template][], with the following template parameters:

| Key                   | Description                                                                                                                          |
|-----------------------|--------------------------------------------------------------------------------------------------------------------------------------|
| `remoteRef`           | The secret ID as specified in the project definition                                                                                 |

If no request body template (`SECRET_PROVIDER_REQUEST_BODY`) has been defined, the webhook service will receive a *HTTP GET* request;
otherwise a *POST* request is issued with the `Content-Type` header set to `application/json; charset=utf-8`. An `Accept` header with
the value `application/json` is included in either case.

An example setup for a webhook service providing secrets via HTTP GET requests:

```bash
SECRET_PROVIDER_SITE_URL="https://secrets.example.com/api/v1/secrets?ref={{ urlquery .remoteRef }}"
```

Alernatively if the service provides secrets for requests with the secret ID in the URL path:

```bash
SECRET_PROVIDER_SITE_URL="https://secrets.example.com/api/v1/secrets/{{.remoteRef}}.json"
```

An example setup for a webhook service providing secrets via HTTP POST requests:

```bash
SECRET_PROVIDER_SITE_URL="https://secrets.example.com/api/v1/secrets"
SECRET_PROVIDER_REQUEST_BODY="{\"secret\":\"{{.remoteRef}}\"}"
```

The response is parsed as *JSON* and the secret value is extracted from the data using a [JMESPath][] query.

Assuming the remote service responds with secret values using the following example, the desired
secret value can be extracted with the [JMESPath][] query `history[-1].value`:

```json
{
  "history": [
    {"changed": "1970-01-01T00:00:00Z", "value": "ChangeMe!", "comment": "initial password"},
    {"changed": "1980-01-01T00:00:00Z", "value": "example", "comment": "changed after leak"},
    {"changed": "1990-01-01T00:00:00Z", "value": "Example123!", "comment": "new and secure"}
  ]
}
```

[JMESPath]: https://jmespath.org/
[Golang Text Template]: https://pkg.go.dev/text/template

kimdre:
Thank you!

Issue #1079, opened 2026-03-06
Deployment fails with "no such file or directory" for compose.yml when using compose.yaml


danroc:
### Description

After upgrading to v0.72.0, all deployments fail with the error:

```text
failed to check for extends file changes: failed to get extends files from compose yaml: open /data/<stack-path>/compose.yml: no such file or directory
```

My compose files are named `compose.yaml`, not `compose.yml`.

### Steps to reproduce the issue

```bash
1. Deploy a stack using compose.yaml on a version prior to v0.72.0
2. Upgrade to v0.72.0
3. Push a commit to trigger a new deployment
4. Deployment fails
```

### Operating System

Linux

### Docker Compose File

```yaml
All my compose files are failing.
```

### Deployment Config

```yaml

```

### Logs

```sh
{"time":"2026-03-06T07:52:09.403929546+01:00","level":"error","msg":"deployment failed","deploy":{"error":"failed to check for extends file changes: failed to get extends files from compose yaml: open /data/<stack-path>/compose.yml: no such file or directory","reference":"refs/heads/main","stack":"<stack-name>"},"job_id":"019cc1ea-fa25-7cbb-8199-74efd70a2892","repository":"<repo>"}
```

danroc:
There is a quick workaround: set the `compose_files` for all stacks in the `.doco-cd.yaml` config file:

```yaml
compose_files:
  - compose.yaml
```

kimdre:
Thank you again for your PR. I released v0.72.1

PR #1080, opened 2026-03-06
fix(compose): skip missing files in extends/include extraction

danroc:
The file list may contain candidate names that don't exist on disk (e.g., `docker-compose.yml` instead of `compose.yml`). Skip these instead of returning an error.

Fixes: #1079


kimdre:
Thank you for the PR. :)

Issue #1086, opened 2026-03-07
If the data volume bind source ends with `/`, doco-cd will create a directory instead of a symlink.


qianlongzt:
### Description

```
{"time":"2026-03-07T20:45:01.804119749+08:00","level":"debug","msg":"retrieved doco-cd data mount point","mount_point":{"destination":"/data","source":"/data/dococd/"}}
```

```
ls -lah /data/dococd/
total 16K
drwxr-xr-x 4 root root 4.0K Mar  7 20:45 .
drwxr-xr-x 5 root root 4.0K Mar  7 14:56 ..
drwxr-xr-x 2 root root 4.0K Mar  7 20:45 dococd
drwxr-xr-x 3 root root 4.0K Mar  7 14:43 git.z-72.com
```

### Steps to reproduce the issue

```bash
1. wsl docker, compose up end with `/` like `/data/dococd/:/data`
2. cd `/data/dococd`


Client:
 Version:           29.2.1
 API version:       1.53
 Go version:        go1.25.6
 Git commit:        a5c7197
 Built:             Mon Feb  2 17:16:41 2026
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Desktop 4.63.0 (220185)
 Engine:
  Version:          29.2.1
  API version:      1.53 (minimum version 1.44)
  Go version:       go1.25.6
  Git commit:       6bc6209
  Built:            Mon Feb  2 17:17:24 2026
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v2.2.1
  GitCommit:        dea7da592f5d1d2b7755e3a161be07f43fad8f75
 runc:
  Version:          1.3.4
  GitCommit:        v1.3.4-0-gd6d73eb8
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

1. I test `/data/dococd:/data` work well.


2. I test on linux both work ok.
```

### Operating System

Other

### Docker Compose File

```yaml
N/A
```

### Deployment Config

```yaml
name: dococd
services:
  app:
    container_name: doco-cd
    build:
      context: .
      dockerfile: Dockerfile
      args:
        - APP_VERSION=dev
    environment:
      HTTP_PORT: "5580"
      LOG_LEVEL: debug
      SOPS_AGE_KEY_FILE: /run/secrets/sops-age-key
      TZ: Asia/Shanghai
    healthcheck:
      test:
        - CMD
        - /doco-cd
        - healthcheck
[...]

PR #1087, opened 2026-03-07
fix: mountPoint source ending with `/`

qianlongzt:
fix: https://github.com/kimdre/doco-cd/issues/1086


kimdre:
Thank you for the fix! :)

Issue #1132, opened 2026-03-17
Unexpected redeploy due to incorrect file change detection


qianlongzt:
### Description

`mihomo` hasn’t changed, but the logs show its config has changed. Maybe it’s because the filename is the same (`config.yaml`)?
```
{"time":"2026-03-17T21:59:04.357901729+08:00","level":"debug","msg":"comparing commits","custom_target":"cwhc-istoreos","deploy":{"deployed_commit":"b7afa60553c8eabf8457fec96d101b1a02974ca6","latest_commit":"5438fba4dfc267c65e8c90b44cfb54d55b63d76b","reference":"refs/heads/master","repository":"git.z-72.com/base/server-notes","stack":"mihomo","stage":"pre-deploy"},"job_id":"019cfc17-ca1c-74af-b20c-e7d0913792d9","repository":"git.z-72.com/base/server-notes"}
{"time":"2026-03-17T21:59:04.448394003+08:00","level":"debug","msg":"changes detected, proceeding with deployment","custom_target":"cwhc-istoreos","deploy":{"directory":"server/cwhc-istoreos/services-auto/mihomo","has_changes":{"compose_config":false,"files":[{"Type":"bindMounts","Services":["mihomo"]}],"images":false},"reference":"refs/heads/master","repository":"git.z-72.com/base/server-notes","stack":"mihomo","stage":"pre-deploy"},"job_id":"019cfc17-ca1c-74af-b20c-e7d0913792d9","repository":"git.z-72.com/base/server-notes"}
{"time":"2026-03-17T21:59:04.448459753+08:00","level":"debug","msg":"completed stage: pre-deploy","custom_target":"cwhc-istoreos","deploy":{"duration":"338ms","reference":"refs/heads/master","repository":"git.z-72.com/base/server-notes","stack":"mihomo","stage":"pre-deploy"},"job_id":"019cfc17-ca1c-74af-b20c-e7d0913792d9","repository":"git.z-72.com/base/server-notes"}
{"time":"2026-03-17T21:59:04.448515981+08:00","level":"debug","msg":"begin stage: deploy","custom_target":"cwhc-istoreos","deploy":{"reference":"refs/heads/master","repository":"git.z-72.com/base/server-notes","stack":"mihomo","stage":"deploy"},"job_id":"019cfc17-ca1c-74af-b20c-e7d0913792d9","repository":"git.z-72.com/base/server-notes"}
[...]

PR #1137, opened 2026-03-18
fix(docker): fix change detect error

qianlongzt:
fix: https://github.com/kimdre/doco-cd/issues/1132


qianlongzt:
@kimdre 
The test failed with `SOPS secret key is not set`. Is that related to a fork PR?

`failed to decrypt project files: failed to decrypt project file '/tmp/TestDeploySwarmStack813920978/001/test.enc.env': failed to decrypt file /tmp/TestDeploySwarmStack813920978/001/test.enc.env: SOPS secret key is not set`

kimdre:
One of the tests did not load the SOPS age key. It's fixed now

kimdre:
Thank you for the PR! :)

PR #1164, opened 2026-03-26
fix(swarm): only redeploy services when changed

### LINKED ISSUE #1153: Only deploy service when changed in swarm mode
### Description

When using swarm mode, changing a single service seems to trigger a redeployment of all services in the stack. I think in Compose mode, only affected services are recreated.

Is this be related to how dynamic labels (like the deployment timestamp) are applied to all services on each deployment? I am happy to do a PR if pointed in the right direction.

### Use Case

I have a stack with 10 services. When I push a change that only affects one service, I'd expect only that service to be redeployed.

### Additional Information

_No response_

zechmeister:
## Summary

Fixes #1153

In swarm mode, every deployment triggered a rolling update on **all** services, even when only one service actually changed. This happened because volatile labels (deployment timestamp, trigger SHA) were placed in container labels (`s.Labels`), which map to `TaskTemplate.ContainerSpec.Labels` in the Swarm ServiceSpec. Any change there causes Docker Swarm to recreate tasks.

This PR moves volatile labels to service-level deploy labels (`s.Deploy.Labels`), which map to `ServiceSpec.Annotations.Labels`. Changes to service-level labels do **not** trigger task recreation.

## Changes

- **`internal/docker/swarm.go`** — Split `addSwarmServiceLabels` into stable container labels and volatile service-level labels
- **`internal/docker/utils.go`** — Add `mergeSwarmLabels` to combine both label sources when reading back; update `GetServiceLabels` and `GetLabeledServices` to use it
- **`internal/docker/swarm_test.go`** — Tests for label placement and stability across redeployments
- **`internal/docker/utils_test.go`** — Test for label merging with precedence

## Test plan

- [x] `TestAddSwarmServiceLabels` — volatile labels in deploy labels, stable in container labels, user labels preserved
- [x] `TestSwarmServiceLabelsStability` — container labels unchanged when only timestamp differs
- [x] `TestMergeSwarmLabels` — service + container labels merged correctly with container precedence
- [x] Full test suite passes (`make test-nobitwarden`)

### DIFF
--- a/internal/docker/swarm.go
+++ b/internal/docker/swarm.go
@@ -80,18 +80,18 @@ func RemoveSwarmStack(ctx context.Context, dockerCli command.Cli, namespace stri
 	return swarmInternal.RunRemove(ctx, dockerCli, opts)
 }
 
-// addSwarmServiceLabels adds custom labels to the service containers in a Docker Swarm stack.
+// addSwarmServiceLabels adds custom labels to services in a Docker Swarm stack.
+// Volatile labels (timestamp, trigger) are placed on service-level deploy labels
+// to avoid altering the task template and triggering unnecessary rolling updates.
 func addSwarmServiceLabels(stack *composetypes.Config, deployConfig *config.DeployConfig, payload *webhook.ParsedPayload,
 	repoDir, appVersion, timestamp, latestCommit, projectHash string,
 ) {
-	customLabels := map[string]string{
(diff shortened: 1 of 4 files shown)


kimdre:
Hi, are you sure your PR is working as intended? When I try to deploy a stack with two services and change the image tag or label of one of them, both still get updated.

PR #1209, opened 2026-04-11
feat(self-update): implement self-update functionality

### LINKED ISSUE #827: Manage doco-cd configuration changes/image updates using doco-cd itself
### Description

It's a common pattern with GitOps and CD to also manage the tool that does the CD... with the tool itself.

For example, it's common practice to manage ArgoCD with ArgoCD itself, after a first stage of bootstrapping.
This feature would allow for a more intuitive and dev-friendly way of managing doco-cd. After all, it _is_ a container, hence it should also be managed the same way other containers are!

### Use Case

I would like to ensure that every configuration change to doco-cd, as well as any update to the image tag, are versioned and controlled like any other deployed stack.

### Additional Information

I have attempted to force this behavior onto doco-cd, but it does not appear to be possible at the moment.

In order to attempt this, I added the following to `.doco-cd.yaml`:

```
---
name: doco-cd
working_dir: services/doco-cd
```

and also gave the following labels to the doco-cd container itself:
 
```
    labels:
      cd.doco.metadata.manager: doco-cd
      cd.doco.repository.name: CapgeminiCisRedTeam/tools

```

These labels allow to bypass [this check](https://github.com/kimdre/doco-cd/blob/031508488738e9b3c9a1d97729c28976d3f5066e/cmd/doco-cd/handler_poll.go#L482) and trick doco-cd into (attempting to) manage itself.

However, the end result was that doco-cd was just erasing its container without recreating it. I haven't had the time to dig deeper in the code, but I assume it's due to how containers are handled (first destroyed and then recreated? Or maybe the doco-cd process is  killed and the container never comes back since the code that should recreate it never runs, not sure).

Moreover, this approach could easily break due to an update on doco-cd internals.

I would gladly try to contribute to this in my free time, though I'm admittedly not very knowledgeable in Go.
Moreover, I've only skimmed the code base (and found a [remarkably long function](https://github.com/kimdre/doco-cd/blob/main/cmd/doco-cd/handler_poll.go#L145) 👀), but I can give this a try if given some pointers!

kimdre:
This pull request introduces a self-update feature for the `doco-cd` application, allowing the container to check for and apply updates to itself both via a new API endpoint and on a scheduled basis. The changes include configuration options, API handler logic, integration with the updater module, and comprehensive tests to ensure correct behavior.

Key changes:

**Self-update feature integration:**

- Added new configuration options to `AppConfig` for enabling self-update, setting the update interval, and controlling health check behavior (`SELF_UPDATE_ENABLED`, `SELF_UPDATE_INTERVAL`, `SELF_UPDATE_WAIT_HEALTHY`). These options are validated and tested in the configuration tests. [[1]](diffhunk://#diff-5eb6005f9d9b182f16db1bb339880faf79c347612932ae5f5a6c62b843005874R40-R42) [[2]](diffhunk://#diff-5af05cfb9d2f38997d40424d0f1ed8ba15b8f66b48c171441c0ed4c625b09357R30-R32) [[3]](diffhunk://#diff-5af05cfb9d2f38997d40424d0f1ed8ba15b8f66b48c171441c0ed4c625b09357R76-R87) [[4]](diffhunk://#diff-5af05cfb9d2f38997d40424d0f1ed8ba15b8f66b48c171441c0ed4c625b09357R156-R169)

- Integrated the updater module throughout the codebase, wiring up the `appUpdater` to the main handler and enabling the self-update scheduler when configured. The updater is initialized in `main.go` and attached to the handler data structure for use by API endpoints. [[1]](diffhunk://#diff-dfbc8c7c81be7e4dfc591f15ed0d387439d6e2040d234cae5bd8d1cab2aabf33R32) [[2]](diffhunk://#diff-dfbc8c7c81be7e4dfc591f15ed0d387439d6e2040d234cae5bd8d1cab2aabf33R288-R308) [[3]](diffhunk://#diff-dfbc8c7c81be7e4dfc591f15ed0d387439d6e2040d234cae5bd8d1cab2aabf33R319-R331) [[4]](diffhunk://#diff-aa9558af838f44fa6bcf58c71d15b8c50d87a208d32a35d119702786114a0d95R33-R46) [[5]](diffhunk://#diff-aa9558af838f44fa6bcf58c71d15b8c50d87a208d32a35d119702786114a0d95R55-R56)

**API endpoint for triggering self-update:**

- Added a new API endpoint `/app/update` (conditionally registered based on configuration) that triggers an asynchronous self-update check. The handler validates the request, ensures only one update runs at a time, and returns appropriate status codes and messages. [[1]](diffhunk://#diff-2e57e32e37823e5f62555afa571e6b27c241df12175fff24b9d6aba379af6bc4R23) [[2]](diffhunk://#diff-2e57e32e37823e5f62555afa571e6b27c241df12175fff24b9d6aba379af6bc4R55-R58) [[3]](diffhunk://#diff-2e57e32e37823e5f62555afa571e6b27c241df12175fff24b9d6aba379af6bc4R789-R834)

**Testing and validation:**

- Implemented thorough tests for the new API endpoint, covering enabled/disabled states and update-in-progress scenarios, as well as configuration parsing and validation for the new options. [[1]](diffhunk://#diff-250ad146be0363ffb5c51944b14e2f2529e17bf197ec0a31c76720a18e6cd76bR24-R37) [[2]](diffhunk://#diff-250ad146be0363ffb5c51944b14e2f2529e17bf197ec0a31c76720a18e6cd76bR331-R418) [[3]](diffhunk://#diff-5af05cfb9d2f38997d40424d0f1ed8ba15b8f66b48c171441c0ed4c625b09357R30-R32) [[4]](diffhunk://#diff-5af05cfb9d2f38997d40424d0f1ed8ba15b8f66b48c171441c0ed4c625b09357R76-R87) [[5]](diffhunk://#diff-5af05cfb9d2f38997d40424d0f1ed8ba15b8f66b48c171441c0ed4c625b09357R156-R169)

- Added a new test file for the updater module to ensure correct behavior of container name normalization, option preservation, network config stripping, and Swarm detection.

**Minor improvements:**

- Improved resource cleanup in `TriggerPollHandler` by using a safer deferred body close pattern.


kimdre:
Current implementation fails because the new cloned container has the same ports as the original container, that triggers the update.
Alternative solution: Create a small sidecar container that handles the update process by
1. Stop old container
2. Create copy of old container with new image version
3. Wait until new container is healthy
4. Remove old container or on healthcheck wait timeout
    1. Remove new container
    2. Start old container again
5. Remove sidecar container

PR #1232, opened 2026-04-19
fix(test): change tag to new tag without docs repo

qianlongzt:


kimdre:
Ah thank you! I did oversee this. I just migrated the old wiki to a new website. Mainly because the old wiki had some annoying limitations but also to make it easier for users to contribute to the documentation. :)

qianlongzt:
The new documentation site is better than the GitHub wiki, mainly because the wiki doesn't support submitting pull requests from forks.

PR #1256, opened 2026-04-23
ci: build doco-cd for riscv64

shanduur:
This PR adds minor change to build riscv64 images on release.

Signed-off-by: Mateusz Urbanek <mateusz.urbanek.98@gmail.com>

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -15,7 +15,7 @@ jobs:
     with:
       build-args: |
         APP_VERSION=${{ github.ref_name }}
-      platforms: linux/amd64,linux/arm64,linux/arm/v7
+      platforms: linux/amd64,linux/arm64,linux/arm/v7,linux/riscv64
       cache: true
       cache-mode: max
       output: image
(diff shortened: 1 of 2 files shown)


kimdre:
Thank you :)

kimdre:
I will need to revert this for now unfortunately. I just tested the build performance and it's very bad when I build it via QEMU since I cannot use runners with riscv hardware.
I'm waiting for a feature request for the build action to allow specifying custom runners so I can use a riscv runner provided by RISE: https://github.com/docker/github-builder/issues/161
https://github.com/apps/rise-risc-v-runners-personal

shanduur:
How slow it was @kimdre? Maybe this could be built only on tags instead of always?

shanduur:
Just for reference, I was building e.g. Kubernetes components for RISC-V, and the difference was negligible, but instead of using QEMU I was using the native build with GOOS/GOARCH which makes cross compilation much faster:

Example run: https://github.com/anza-labs/docker-library/actions/runs/23082795348

Dockerfile: https://github.com/anza-labs/docker-library/blob/main/library/kubernetes/Dockerfile

kimdre:
> How slow it was @kimdre? Maybe this could be built only on tags instead of always?

Almost one hour. This is way too long when I have to release hotfixes.

shanduur:
Oh wow, when I was building it locally it was around 4 minutes…

PR #1268, opened 2026-04-26
feat: trigger reconciliation on Docker daemon events

### LINKED ISSUE #1249: Run reconciliation by listening for Docker daemon events
### Description

This is a follow-up to #1208

Subscribe to the Docker API endpoint `/events` and listen for container removal events in the background to trigger the reconciliation only when necessary without having to use an interval.

### Use Case

The current reconciliation implementation needs to run at an interval to check for pending tasks even when there is nothing to do. Listening on the /events endpoint reduces noise in the application logs.

### Additional Information

Examples for listening for container stop and start events:

```sh
curl --unix-socket /var/run/docker.sock "http://localhost/events?filters=%7B%22type%22%3A%5B%22container%22%5D%2C%22action%22%3A%5B%22start%22%2C%22destroy%22%5D%7D"
```

### Go

```sh
go mod init docker-events
go get github.com/docker/docker/client
```

```python
package main

import (
	"context"
	"encoding/json"
	"fmt"
	"log"

	"github.com/docker/docker/api/types"
	"github.com/docker/docker/api/types/events"
	"github.com/docker/docker/api/types/filters"
	"github.com/docker/docker/client"
)

func main() {
	ctx := context.Background()

	// Initialize client
	cli, err := client.NewClientWithOpts(client.FromEnv)
	if err != nil {
		log.Fatalf("Failed to create client: %v", err)
	}

	// Set filters: Type 'container' AND Actions 'die' OR 'destroy'
	// Note: 'die' covers stopping, killing, and crashing.
	filterArgs := filters.NewArgs()
	filterArgs.Add("type", "container")
	filterArgs.Add("action", "die")
	filterArgs.Add("action", "destroy")

	options := types.EventsOptions{
		Filters: filterArgs,
	}

	fmt.Println("Listening for container 'die' (stopped) and 'destroy' (removed) events...")

	// Connect to the Events API
	stream, err := cli.Events(ctx, options)
	if err != nil {
		log.Fatalf("Failed to subscribe to events: %v", err)
	}
	defer stream.Close()

	decoder := json.NewDecoder(stream)

	for {
		var event events.Message
		if err := decoder.Decode(&event); err != nil {
			log.Printf("Error decoding event stream: %v", err)
			break
		}

		fmt.Printf(
			"Event: Action=%s | Container ID=%s\n",
			event.Action,
			event.Actor.ID[:12],
		)
	}
}
```

kimdre:
This PR changes the reconciliation logic from a job that runs on an interval to a event-driven check.
Doco-CD listents on the Docker Daemon `/events` API endpoint for new events.
When configured container events occur, doco-cd automatically reapplies the deployment to bring the stack back to the desired state.

The following settings can be used to configure reconciliation triggers.

| Key       | Type             | Description                                                                                                   | Default value        |
|-----------|------------------|---------------------------------------------------------------------------------------------------------------|----------------------|
| `enabled` | boolean          | Enable reconciliation.                                                                                        | `true`               |
| `events`  | array of strings | Docker container/service events that trigger reconciliation. See [supported events](#supported-events) below. | `['die', 'unhealthy']` |

> [!NOTE]
> [Reconciliation](https://doco.cd/latest/Deploy-Settings/#reconciliation-settings) for non-Swarm deployments follows classic Compose `restart` semantics:
>
>    - Services with `restart: always` or `restart: unless-stopped` are expected to stay running.
>    - Services with no explicit `restart` policy are treated as `restart: "no"`.
>    - Services with `restart: on-failure` may remain exited after success (exited with status code `0`), and `restart: "no"` is treated as one-time behavior and is not reconciled back to running.
> - Restart events can override the stop signal using `reconciliation.restart_signal`.
>  - To prevent endless restart loops caused by flappy health checks, ` unhealthy` restarts are rate-limited via `reconciliation.restart_limit` and `reconciliation.restart_window`.
>
>  More information on the restart policies can be found in the [Docker Compose specification](https://docs.docker.com/reference/compose-file/services/#restart).

>  [!IMPORTANT] 
> **Reconciliation in Docker Swarm**
> Docker Swarm manages some desired-state reconciliation by itself with Swarm service modes and [`deploy.restart_policy`](https://docs.docker.com/reference/compose-file/deploy/#restart_policy) behavior. See 
> Docker's documentation on [desired state reconciliation](https://docs.docker.com/engine/swarm/#desired-state-reconciliation).  
> Doco-CD's reconciliation for Swarm deployments only manages service updates and scaling, but not container restarts or health status.

### Supported Events

Events can be triggered by changes in the container state, configuration updates outside Doco-CD (e.g. via Docker CLI), or health status changes.
The following events are supported as reconciliation triggers in Docker (Standalone) and Docker Swarm deployments:

| Event       | Description                                                            | Standalone | Swarm | Action   |
|-------------|------------------------------------------------------------------------|------------|-------|----------|
| `die`       | The container process exited.                                          | Yes        | No    | Redeploy |
| `destroy`   | The container was removed / service was removed.                       | Yes        | Yes   | Redeploy |
| `update`    | The service/container configuration was updated (for example scaling). | No         | Yes   | Redeploy |
| `stop`      | The container was stopped gracefully.                                  | Yes        | No    | Restart  |
| `kill`      | The container was terminated by a signal.                              | Yes        | No    | Restart  |
| `oom`       | The container was killed because it ran out of memory.                 | Yes        | No    | Restart  |
| `unhealthy` | The container health check status changed to _unhealthy_.              | Yes        | No    | Restart  |

> [!NOTE]
> **Flapping health checks**
> For `unhealthy` events, doco-cd suppresses further automatic restarts after `restart_limit` restarts within `restart_window` seconds.

### Examples

```yaml
name: some-project
reconciliation:
  enabled: true
  restart_timeout: 30
  restart_signal: SIGQUIT
  restart_limit: 5
  restart_window: 300
  events:
    - die
    - stop
    - kill
    - unhealthy
    - oom
```

```yaml
name: some-project
reconciliation:
  enabled: true
  events: [die, destroy]
```


kimdre:
@qianlongzt What do you think about my change?

qianlongzt:
@kimdre For Docker Swarm, should we watch delete and update events in case users delete or update services?

kimdre:
@qianlongzt I added support for destroy and update events in Swarm Mode. The rest should already get handled by Docker Swarm itsef.

internal/config/deploy_config.go
```diff
@@ -156,6 +188,49 @@ func (c *DeployConfig) validateConfig() error {
...
+		c.Reconciliation.Events = append([]string(nil), defaultReconciliationEvents...)
+		return nil
+	}
+
+	normalized := make([]string, 0, len(c.Reconciliation.Events))
+	seen := make(map[string]struct{}, len(c.Reconciliation.Events))
+
+	for _, rawEvent := range c.Reconciliation.Events {
+		event := strings.ToLower(strings.TrimSpace(rawEvent))
+		event = strings.Join(strings.Fields(event), " ")
```
qianlongzt: `event = strings.Join(strings.Fields(event), " ")`
useless ?

wiki/docs/Deploy-Settings.md
```diff
@@ -217,28 +217,65 @@ destroy_opts:
...
+### Reconciliation Settings
 
-Reconciliation is an optional periodic check that compares the currently running services with the expected deployment state.
-If drift is detected, doco-cd automatically reapplies the deployment to bring the stack back to the desired state.
+Reconciliation is an optional event-driven check that compares the currently running Docker services/containers with the expected deployment state.
+When configured container events occur, doco-cd either reapplies the deployment or directly restarts the affected container, depending on the event type.
 
-The following settings can be used to configure periodic reconciliation.
-
-!!! warning
```
qianlongzt, line 228: The state will still be lost after a restart.
Maybe we should clarify that it will be restored on the next poll or webhook event.

internal/config/deploy_config.go
```diff
@@ -73,8 +82,12 @@ type DeployConfig struct {
...
 	} `yaml:"auto_discover_opts" json:"auto_discover_opts"` // AutoDiscoverOpts are options for the autodiscovery feature
 	Reconciliation struct {
-		Enabled  bool `yaml:"enabled" json:"enabled" default:"true"` // Enabled enables the reconciliation feature
-		Interval int  `yaml:"interval" json:"interval" default:"60"` // Interval is the interval in seconds at which the reconciliation job is run
+		Enabled        bool     `yaml:"enabled" json:"enabled" default:"true"`                   // Enabled enables the reconciliation feature
+		Events         []string `yaml:"events" json:"events" default:"[\"die\", \"unhealthy\"]"` // Events is the list of Docker container actions that trigger reconciliation
+		RestartTimeout int      `yaml:"restart_timeout" json:"restart_timeout" default:"10"`     // RestartTimeout is the timeout in seconds to wait before killing a container during a restart (used for restart-oriented reconciliation events)
+		RestartSignal  string   `yaml:"restart_signal" json:"restart_signal" default:"SIGTERM"`  // RestartSignal is the signal sent to stop containers during a restart. Empty uses Docker's default signal.
```
qianlongzt: Is the default empty?

The default is SIGTERM. Users need to set it manually to restore the default behavior.

internal/reconciliation/reconciliation.go
```diff
@@ -69,57 +88,278 @@ func (j *job) close() {
...
+				return true // reconnect
+			}
+
+			j.handleEvent(ctx, jobLog, event)
+		}
+	}
+}
+
+// normalizeRepositoryLabelFromCloneURL mirrors the repository-name normalization used during poll deployment labeling.
+func normalizeRepositoryLabelFromCloneURL(cloneURL config.HttpUrl) string {
```
qianlongzt: replace with git.GetFullName

internal/reconciliation/reconciliation.go
```diff
@@ -69,57 +88,278 @@ func (j *job) close() {
...
+	}
+
+	if len(parts) == 2 {
+		return parts[1]
+	}
+
+	return repoName
+}
+
+func dockerEventTypeForMode(swarmMode bool) string {
```
qianlongzt: Maybe we should split event-related functions into a separate file?
A large file (~700 lines) is hard to review.

internal/logger/attrs.go
```diff
@@ -0,0 +1,135 @@
...
+		filteredAttrs = append(filteredAttrs, attr)
+	}
+
+	return &attrFilterHandler{
+		next:  h.next,
+		attrs: filteredAttrs,
+	}
+}
+
+type recordAttrFilterHandler struct {
```
qianlongzt, line 78: What’s the difference between this and `attrFilterHandler`? Can they be merged?

internal/reconciliation/reconciliation.go
```diff
@@ -45,16 +54,26 @@ type jobInfo struct {
 type job struct {
 	info jobInfo
 
-	// key is the interval in second
-	deployConfigGroupByInterval map[int][]*config.DeployConfig
-	closeChan                   chan struct{}
+	// key is the docker event action name (for example "die" or "unhealthy").
+	deployConfigGroupByEvent map[string][]*config.DeployConfig
+	unhealthyRestartHistory  map[string][]time.Time
+	restartSuppressUntil     map[string]time.Time
```
qianlongzt: A comment explaining what the key represents would be helpful.

internal/reconciliation/reconciliation.go
```diff
@@ -69,57 +88,278 @@ func (j *job) close() {
...
+				continue
+			}
+
+			filters["destroy"] = struct{}{}
+		default:
+			filters[action] = struct{}{}
+		}
+	}
+
+	return mapsKeys(filters)
```
qianlongzt: We can reuse `github.com/kimdre/doco-cd/internal/utils/set`.

internal/reconciliation/reconciliation.go
```diff
@@ -69,57 +88,278 @@ func (j *job) close() {
...
+		}
+	}
+
+	return result
+}
+
+func stackNameFromEvent(event events.Message, candidates []*config.DeployConfig) string {
+	attrs := event.Actor.Attributes
+
+	for _, key := range []string{docker.DocoCDLabels.Deployment.Name, swarm.StackNamespaceLabel} {
```
qianlongzt: Maybe we should include the `com.docker.compose.project` label here?

internal/reconciliation/reconciliation.go
```diff
@@ -69,57 +88,278 @@ func (j *job) close() {
...
+	if serviceName == "" {
+		serviceName = strings.TrimSpace(attrs["service"])
+	}
+
+	if serviceName == "" {
+		return ""
+	}
+
+	for _, dc := range candidates {
+		if serviceName == dc.Name || strings.HasPrefix(serviceName, dc.Name+"_") {
```
qianlongzt: In which cases do we need to get the stack name from `name` or `service`?
Should the separator be `-`?

internal/reconciliation/reconciliation.go
```diff
@@ -128,24 +368,358 @@ func (j *job) deploy(ctx context.Context, jobLog *slog.Logger, dcs []*config.Dep
...
+	for _, dc := range dcs {
+		if dc != nil {
+			return dc
+		}
+	}
+
+	return nil
+}
+
+func (j *job) restartUnhealthyContainersOnStartup(ctx context.Context, jobLog *slog.Logger) {
```
qianlongzt: Maybe we need to check for missing services on startup?

wiki/docs/Deploy-Settings.md
kimdre: I readded the banner and made it more clear.

internal/reconciliation/reconciliation.go
kimdre: I added some comments to explain the key

internal/logger/attrs.go
kimdre: Technically yes, but I'd like to keep  them seperate because they serve different purposes.

`attrFilterHandler`: captures/reapplies logger-level attrs and supports `WithoutAttr` removal of attached attrs.
`recordAttrFilterHandler`: removes one key from emitted records when the underlying handler is a unknown implemenation (not `attrFilterHandler)`.

internal/reconciliation/reconciliation.go
kimdre: Thanks for the hint, I will implement that.

internal/reconciliation/reconciliation.go
kimdre: Yes you are right. I will do that.

kimdre:
@qianlongzt I think the PR is ready now. Do you want to do some testing before I merge it?

qianlongzt:
I will review it today.

internal/reconciliation/reconciliation.go
```diff
@@ -2,197 +2,302 @@ package reconciliation
...
-	wg.Wait()
-}
+		listenerCtx, cancel := context.WithCancel(ctx)
 
-func (j *job) runByInterval(ctx context.Context, interval int, dcs []*config.DeployConfig) {
-	if len(dcs) == 0 {
-		return
-	}
+		eventResult := j.info.dockerCli.Client().Events(listenerCtx, client.EventsListOptions{
+			Filters: filterArgs,
```
qianlongzt, line 89: Add a since filter to avoid missing events after startup recovery.

internal/reconciliation/reconciliation.go
```diff
@@ -2,197 +2,302 @@ package reconciliation
...
+	// so reconciliation can recover them without waiting for a new health event.
+	// Run the one-time startup recovery checks in parallel, but wait for both to
+	// finish before subscribing to Docker events so all startup healing happens
+	// against a stable initial view of the daemon state.
+	var startupRecoveryWG sync.WaitGroup
+	startupRecoveryWG.Add(2)
 
-func (j *job) run(ctx context.Context) {
-	jobLog := j.info.jobLog
+	go func() {
```
qianlongzt: Use `wg.Go` to avoid manually calling `wg.Add` and `wg.Done`.

internal/reconciliation/reconciliation_helpers.go
```diff
@@ -0,0 +1,230 @@
...
+	for _, key := range []string{"name", "service", "com.docker.swarm.service.name", "com.docker.swarm.task.name"} {
+		identifier := strings.TrimSpace(attrs[key])
+		if identifier == "" {
+			continue
+		}
+
+		if matched := matchCandidateStackName(identifier, candidates); matched != "" {
+			return matched
+		}
+	}
```
qianlongzt: The two loops seem the same. Can they be merged?

internal/reconciliation/reconciliation_event_test.go
```diff
@@ -0,0 +1,682 @@
...
+	"github.com/moby/moby/api/types/events"
+
+	"github.com/kimdre/doco-cd/internal/config"
+	"github.com/kimdre/doco-cd/internal/docker"
+	"github.com/kimdre/doco-cd/internal/docker/swarm"
+	gitInternal "github.com/kimdre/doco-cd/internal/git"
+	"github.com/kimdre/doco-cd/internal/notification"
+)
+
+func TestNormalizeRepositoryLabelFromCloneURL(t *testing.T) {
```
qianlongzt: This test should be merged into `internal/git.TestGetFullName`

internal/reconciliation/reconciliation_helpers.go
```diff
@@ -0,0 +1,230 @@
...
+	return result
+}
+
+// stackNameFromEvent attempts to determine the stack name referenced by the given Docker event
+// by examining various event attributes and matching them against the candidate config.DeployConfig configs.
+// Returns an empty string when no stack name could be determined or matched.
+func stackNameFromEvent(event events.Message, candidates []*config.DeployConfig) string {
+	attrs := event.Actor.Attributes
+
+	for _, key := range []string{docker.DocoCDLabels.Deployment.Name, swarm.StackNamespaceLabel} {
```
qianlongzt: Maybe we should include the `com.docker.compose.project` label here?

internal/config/deploy_config.go
```diff
@@ -73,8 +82,12 @@ type DeployConfig struct {
...
 	} `yaml:"auto_discover_opts" json:"auto_discover_opts"` // AutoDiscoverOpts are options for the autodiscovery feature
 	Reconciliation struct {
-		Enabled  bool `yaml:"enabled" json:"enabled" default:"true"` // Enabled enables the reconciliation feature
-		Interval int  `yaml:"interval" json:"interval" default:"60"` // Interval is the interval in seconds at which the reconciliation job is run
+		Enabled        bool     `yaml:"enabled" json:"enabled" default:"true"`                   // Enabled enables the reconciliation feature
+		Events         []string `yaml:"events" json:"events" default:"[\"die\", \"unhealthy\"]"` // Events is the list of Docker container actions that trigger reconciliation
+		RestartTimeout int      `yaml:"restart_timeout" json:"restart_timeout" default:"10"`     // RestartTimeout is the timeout in seconds to wait before killing a container during a restart (used for restart-oriented reconciliation events)
+		RestartSignal  string   `yaml:"restart_signal" json:"restart_signal" default:"SIGTERM"`  // RestartSignal is the signal sent to stop containers during a restart.
```
qianlongzt: Is the default empty?

The default is SIGTERM. Users need to set it manually to restore the default behavior.

kimdre: The default stop signal in Docker CLI/client is `SIGTERM` before it sends a `SIGKILL` on timeout. I just added it here to make it more clear that we also use the default signal.

internal/reconciliation/reconciliation_helpers.go
kimdre: The `com.docker.compose.project` label is not used in Swarm stacks. Instead, containers and services have the value of `swarm.StackNamespaceLabel` applied to distinguish between stacks.

internal/reconciliation/reconciliation_event_test.go
kimdre: I removed the test.

internal/reconciliation/reconciliation_helpers.go
kimdre: I merged them

internal/reconciliation/reconciliation_helpers.go
kimdre: I'm not sure if we even need the `swarm.StackNamespaceLabel` since we only want to check for containers and services that were deployed by doco-cd.

qianlongzt:
> The default signal to use is defined by the image's [StopSignal](https://github.com/opencontainers/image-spec/blob/v1.1.0/config.md), which can be set through the [STOPSIGNAL](https://docs.docker.com/reference/dockerfile/#stopsignal) Dockerfile instruction when building the image, or configured using the [--stop-signal](https://docs.docker.com/reference/cli/docker/container/run/#stop-signal) option when creating the container. If no signal is configured for the container, SIGTERM is used as default.

https://docs.docker.com/reference/cli/docker/container/stop/

kimdre:
> > The default signal to use is defined by the image's [StopSignal](https://github.com/opencontainers/image-spec/blob/v1.1.0/config.md), which can be set through the [STOPSIGNAL](https://docs.docker.com/reference/dockerfile/#stopsignal) Dockerfile instruction when building the image, or configured using the [--stop-signal](https://docs.docker.com/reference/cli/docker/container/run/#stop-signal) option when creating the container. If no signal is configured for the container, SIGTERM is used as default.
> 
> https://docs.docker.com/reference/cli/docker/container/stop/

Maybe we can get the required stop signal from the image spec and use `SIGTERM` as fallback?

internal/reconciliation/reconciliation_helpers.go
qianlongzt: `com.docker.compose.project` might be a good fallback for `docker.DocoCDLabels.Deployment.Name`, for stacks not deployed by doco-cd.
Should doco-cd handle this fallback logic?

qianlongzt:
I’m not sure if the Docker daemon already handles this. The API signal not be required.

https://docs.docker.com/reference/api/engine/version/v1.53/#tag/Container/operation/ContainerRestart

qianlongzt:
https://github.com/moby/moby/blob/master/daemon/stop.go#L48
I think the Docker daemon already handles this.

kimdre:
It's not required. As I said the docker cli already has a default value for stopping and killing containers but I want to allow users to set this for graceful stops if they need to.

kimdre:
RestartSignal is now empty/unset by default in deploy config.
If RestartSignal not set, doco-cd attempts to fetch the image's StopSignal; if that is not available, SIGINT is used.

internal/reconciliation/reconciliation_helpers.go
kimdre: Should we allow reconciliation for stacks that were not deployed by doco-cd? I don't think this is a good idea tbh.

qianlongzt: OK, let’s not add the `com.docker.compose.project label`.

internal/reconciliation/reconciliation_restart.go
```diff
@@ -207,24 +237,62 @@ func (j *job) shouldSuppressUnhealthyRestart(jobLog *slog.Logger, event events.M
...
+		signal = strings.TrimSpace(dc.Reconciliation.RestartSignal)
+	}
+
+	if signal == "" && containerID != "" {
+		signal = getImageStopSignalForContainer(ctx, jobLog, cli, containerID)
+	}
+
+	if signal == "" {
+		signal = "SIGINT"
 	}
```
qianlongzt: I don’t think we need `getImageStopSignalForContainer`, since the daemon already handles this logic.

Just simplify it to `opts.Signal = signal`.

https://github.com/moby/moby/blob/master/daemon/restart.go#L63
https://github.com/moby/moby/blob/master/daemon/stop.go#L48
https://github.com/moby/moby/blob/master/daemon/container/container.go#L596

internal/reconciliation/reconciliation_restart.go
```diff
@@ -207,24 +237,62 @@ func (j *job) shouldSuppressUnhealthyRestart(jobLog *slog.Logger, event events.M
...
 	timeout := 10
 	if dc != nil && dc.Reconciliation.RestartTimeout > 0 {
 		timeout = dc.Reconciliation.RestartTimeout
 	}
 
 	opts := client.ContainerRestartOptions{Timeout: &timeout}
 
+	signal := ""
+
+	// Priority: user-configured signal → image StopSignal → default SIGINT
```
qianlongzt: The default in Docker is `SIGTERM`.

internal/reconciliation/reconciliation_restart.go
kimdre: I think you are right. I will revert that

qianlongzt (approved):

PR #1269, opened 2026-04-26
fix(docker): stop false-positive "service not deployed" loop for containers missing cd.doco.* labels

JojoMee:
## Problem

When a container belonging to a doco-cd-managed stack is missing `cd.doco.*` labels, doco-cd enters an infinite redeployment loop on every reconciliation cycle. I encountered this problem when I manually restarted the postgres container in one of my compose stacks to perform a major upgrade. This was before the v0.82.0 update which added periodic reconciliation. After updating to v0.82.0 I got a notification every 60s, that changes were detected and the stack needs to be redeployed.

This happens because `getLatestServiceStatus` uses the `cd.doco.repository.name` label as a gate for **both** concerns:

1. Selecting deployment metadata (latest commit SHA, compose hash)
2. Building the running service inventory (`DeployedStatus`)

Containers without the label are silently excluded from `DeployedStatus`. `CheckServiceMismatch` then reports them as `"service not deployed"`, which causes `shouldSkipDeployment` to return `false`. Since the container spec has not actually changed, Docker Compose does not recreate the container, so the labels are never stamped and the loop continues indefinitely.

## How to reproduce

1. Have a stack managed by doco-cd with `restart: always` or `restart: unless-stopped` services.
2. Recreate one of its containers outside of doco-cd (e.g. `docker compose up -d <service>` run directly on the host, or `docker rm -f <container>` followed by a manual `docker compose up`).
3. Make sure no compose spec change exists that would cause doco-cd to recreate the container on its own.
4. Enable the reconciliation interval (`reconciliation.interval > 0`).
5. Observe doco-cd attempting a redeployment on every poll cycle.

## Fix

Decouple the two concerns in `getLatestServiceStatus`:

- **`DeployedStatus`**: populated from *all* containers returned by `getDeployStatus` (already scoped to the project via `com.docker.compose.project=<name>`). No `cd.doco.*` label required.
- **Metadata** (commit SHA, compose hash): still selected only from containers that carry a matching `cd.doco.repository.name` label, preserving the existing semantics.

I'm not 100% sure that this proposed fix is the ideal solution for the observed problem, one could argue that when doing gitops you're simply not supposed to manually mess with the containers. Maybe you have a better/different solution for this, then feel free to close this PR.

---
> Disclaimer: The root cause analysis, fix, and unit tests were developed with AI assistance. I manually reviewed and validated the fix in my real deployment environment to confirm it resolves the issue.

### DIFF
--- a/internal/docker/service_utils.go
+++ b/internal/docker/service_utils.go
@@ -69,6 +69,14 @@ func getLatestServiceStatus(cacheMap *sync.Map, statusMap map[Service]ServiceSta
 	)
 
 	for serviceName, state := range statusMap {
+		// Always include the service in the deployed inventory, regardless of whether it has
+		// cd.doco.* labels. Containers that are missing those labels (e.g. started via the
+		// Docker CLI directly or deployed before doco-cd stamped them) are still genuinely
+		// running and must not be reported as "service not deployed" by CheckServiceMismatch.
+		ret.DeployedStatus[serviceName] = state
+
+		// Only use containers that carry doco-cd repository labels for deployment metadata
+		// (latest commit SHA, compose hash). This keeps the two concerns separate.
 		labels := state.Labels
 
 		name, ok := labels[DocoCDLabels.Repository.Name]
@@ -86,8 +94,6 @@ func getLatestServiceStatus(cacheMap *sync.Map, statusMap map[Service]ServiceSta
 			latestTimestamp = timestamp
 			latestLabels = labels
 		}
-
-		ret.DeployedStatus[serviceName] = state
 	}
 
 	cache, ok := getDeployStatusFromCache(cacheMap, git.GetRepoName(cloneURL), deployName)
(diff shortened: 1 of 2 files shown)


qianlongzt (approved):

qianlongzt:
This PR looks good to me. The repo label filter seems unnecessary.

1. If a resource is created manually and added to the repo, doco-cd should handle it.
2. If a resource is created manually and added to the repo but should not be managed by doco-cd, not in `.doco-cd.*.yaml`
3. For stack/project conflicts, users will always need to make adjustments.

qianlongzt:
@JojoMee Version [v0.83.0-rc.1](https://github.com/kimdre/doco-cd/releases/tag/v0.83.0-rc.1)
 should fix many notification issues, but this PR still makes sense.

kimdre:
Thanks for the PR! :)

PR #1300, opened 2026-05-04
fix(reconciliation): change default trigger events for reconciliation

kimdre:
Changed the default trigger events for reconciliation from `[die, unhealthy]` to  `[unhealthy]`.


qianlongzt:
Why was the `die` event removed?

If so, `docker rm` or `docker stop` will no longer trigger reconciliation for missing or stopped containers.

kimdre:
> Why was the `die` event removed?
> 
> If so, `docker rm` or `docker stop` will no longer trigger reconciliation for missing or stopped containers.

Because the `die` event is very broad and I would like to reflect the auto-heal behaviour of Docker Swarm services to Standalone containers as the default since this is probaly the most useful scenario for most users (crashes get already handled by the Docker daemon when a `restart` policy is set and I don't want to cause too much confusion for users).
If users want to always let doco-cd recreate their containers they can enable this manually but Docker does not permanently remove containers by itself without a user-interaction.

PR #1311, opened 2026-05-06
Update 'remove_dir' setting description for clarity

mikeage:
Clarified the description for the 'remove_dir' setting to specify use cases involving relative paths and multiple services.

I got bit by this myself, and after reading the docs carefully and thinking about the implications, I thought it might helpful to have a few more words explaining when one would *not* want to delete the directory upon destroying a deployment


kimdre:
Thank you, this will definitely be helpful to other users. :)

PR #1340, opened 2026-05-12
feat(build): vendor buildx CLI plugin to enable BuildKit Dockerfiles

SamORichards:
## Summary

Doco-CD's distroless release image ships only the `/doco-cd` binary, so the Docker CLI's plugin manager never finds buildx on disk. Compose v5 uses that lookup to pick its builder — [`pkg/compose/build.go`](https://github.com/docker/compose/blob/main/pkg/compose/build.go):

```go
bake, err := buildWithBake(s.dockerCli)
if bake { return s.doBuildBake(...) }
return s.doBuildClassic(...)
```

`buildWithBake` calls `manager.GetPlugin("buildx", dockerCli, ...)` and on `errdefs.IsNotFound` falls through to `doBuildClassic`, which POSTs to the daemon's legacy `/build` endpoint. That endpoint uses the pre-BuildKit Dockerfile parser, so any Dockerfile that uses BuildKit-only syntax fails with:

```
the --mount option requires BuildKit
```

Affected syntax includes `RUN --mount=type=cache`, `--mount=type=secret`, `--mount=type=bind` outside the build context, additional build contexts, and so on. Every user Dockerfile is silently restricted to legacy-builder syntax with no in-tool workaround — neither the daemon's `features.buildkit` flag nor `DOCKER_BUILDKIT=1` in the container env change the outcome, because the decision is taken client-side by the compose lib before any daemon call.

This PR vendors `docker/buildx-bin` into the release stage at `/usr/local/libexec/docker/cli-plugins/docker-buildx` (one of the standard Docker CLI plugin search paths). Compose's `manager.GetPlugin` discovers it automatically; no Go change.

## Diff

Four lines added, none changed:

```diff
+# buildx plugin so compose v5 picks BuildKit instead of the legacy ``/build`` endpoint
+FROM docker/buildx-bin:0.33.0@sha256:450be95fa632a3986797cd23b8b5d8d5fff47e9fd8e1fa483c9d44b07da2a559 AS buildx
 ...
 COPY --from=build /doco-cd /doco-cd
+COPY --from=buildx /buildx /usr/local/libexec/docker/cli-plugins/docker-buildx
```

The pinned digest is the manifest-list digest, so each `TARGETARCH` resolves to the matching `buildx` binary at build time.

## Costs

- **Image size** grows ~64 MB (the statically-linked `buildx` binary). The release stage was ~30 MB previously, so this roughly doubles the image.
- **Multi-arch:** `docker/buildx-bin` ships matching variants for amd64, arm64, arm/v7, arm/v6, s390x, riscv64, and ppc64le, so doco-cd's existing build matrix keeps working as-is.

## Verification

- Image builds cleanly: `docker build --target release .`
- Binary lands at `/usr/local/libexec/docker/cli-plugins/docker-buildx`, mode `-rwxr-xr-x`, statically-linked ELF.
- Responds correctly to the CLI plugin metadata handshake — `docker-buildx docker-cli-plugin-metadata` returns valid `SchemaVersion 0.1.0` JSON — so the plugin manager will register it.

Happy to add an integration test (deploy a tiny stack with a `RUN --mount=type=cache,...` Dockerfile and assert build success) if you'd like that as part of the PR.

## Risk to existing users

Legacy Dockerfiles continue to work; bake/buildx supports a strict superset of the classic builder. The only regression surface is users on Docker daemons predating BuildKit (Moby < 18.09, released 2018) — vanishingly rare in 2026.

---

Context: I hit this in a SaaS deploy when adding `RUN --mount=type=cache` to speed up `pip install` and `npm ci`. The error message points at BuildKit but there's no docs path to enable it through doco-cd, since the plugin discovery is internal to compose. Hopefully this is a useful out-of-the-box default for anyone else who lands here.


kimdre:
Can you please show me a example Dockerfile where buildx is required?

SamORichards:
> Can you please show me a example Dockerfile where buildx is required?

Here's the minimal reproducer:

Dockerfile

```
# syntax=docker/dockerfile:1
FROM python:3.13-slim
RUN --mount=type=cache,target=/root/.cache/pip pip install requests
```


docker-compose.yaml
```
services:
  app:
    build: .
```


Build it through any tool that doesn't have buildx on its CLI plugin path:

```
$ DOCKER_BUILDKIT=0 docker compose build
=> ERROR [internal] load build definition from Dockerfile
------
 > [internal] load build definition from Dockerfile:
------
Dockerfile:3
--------------------
   1 |     # syntax=docker/dockerfile:1
   2 |     FROM python:3.13-slim
   3 | >>> RUN --mount=type=cache,target=/root/.cache/pip pip install requests
--------------------
ERROR: failed to solve: the --mount option requires BuildKit. Refer to https://docs.docker.com/go/buildkit/ to learn how to build images with BuildKit enabled
```

DOCKER_BUILDKIT=0 here forces the same code path doco-cd currently takes

kimdre:
Thanks, makes sense to add it to the image :)

kimdre (approved):

PR #1374, opened 2026-05-20
refactor(secretprovider)!: extract providers as gRPC plugins

### LINKED ISSUE #1369: Refactor `secretproviders` to plugins.
### Description

Refactor the current secretproviders implementation to use a gRPC-based plugin architecture instead of in-process or tightly coupled integrations. This would standardize how secret providers are implemented, loaded, and communicated with, enabling external binaries/services to act as providers through a well-defined RPC interface.
The goal is to decouple secret retrieval logic from the core system and move toward a pluggable, versioned API boundary using gRPC.

### Use Case

In real deployments, secrets can come from lots of places like Vault, OpenBao, cloud secret managers, HSMs, or internal systems. Baking all of that into the core quickly makes it complex and tightly coupled, and every new integration slows things down.

A gRPC plugin model avoids that by letting secret providers run separately from the core. You can add or update integrations without rebuilding the main binary, and run them as sidecars or separate services. It also makes it easy to swap backends per environment, like local files in dev and Vault or OpenBao in prod.

It also improves isolation and security since each provider can run with limited permissions, reducing blast radius. And the core stays smaller and simpler because it doesn’t need to ship all the SDKs and clients, which helps with build times and deployments, especially in edge or constrained setups.

### Additional Information

A gRPC plugin system fits well, where the core just defines the contract and external components do the work. gRPC streaming could also help for cases like secret rotation or long-lived leases (e.g. Vault watches). Plugins could be discovered via Unix sockets or config, depending on how it’s deployed.

Alternatively, WASM plugins (e.g. [https://github.com/knqyf263/go-plugin](https://github.com/knqyf263/go-plugin)) could be considered.

shanduur:
Move 1Password, AWS SM, Bitwarden SM, Infisical, and OpenBao out of
the core binary into standalone gRPC plugins so secret backends can be
versioned, swapped, and run with reduced privilege without rebuilding
doco-cd. Core now talks to plugins over the SecretProvider RPC defined
under proto/, generated with buf.

- gRPC contract in proto/secretprovider/v1, generated to api/
- Plugin binaries under cmd/secretproviders/<name> sharing a server harness
- Core dispatches legacy provider names through the gRPC client
- Per-plugin Dockerfile using tonistiigi/xx; core Dockerfile drops CGO
- Matrix workflows build plugin images on tag and dev branches
- Makefile: plugin list as variable, configurable DOCKER_PLATFORMS,
  buf targets

BREAKING CHANGE: Selecting any of 1password, aws_sm, bitwarden_sm,
infisical, openbao via SECRET_PROVIDER now requires a running plugin
reachable at SECRET_PROVIDER_GRPC_ENDPOINT. The nobitwarden build tag
and bundled in-process providers are removed.

Closes #1369

Signed-off-by: Mateusz Urbanek <mateusz.urbanek.98@gmail.com>


kimdre:
Hi thanks for the PR! I didn't check everything yet but can you please update the CONTRIBUTING.md about how to manage/add secret providers with your approach and how to use the the proto generator?

.github/workflows/test.yaml
```diff
@@ -80,26 +80,74 @@ jobs:
...
-      - run: go test -ldflags="$BW_SDK_BUILD_FLAGS" -v -coverprofile=coverage-${{ matrix.mode }}.out -covermode=atomic -timeout 20m ./...
+      - name: Test core + pure-Go plugins
+        run: |
+          go test -v \
+            -coverprofile=coverage-core-${{ matrix.mode }}.out \
+            -covermode=atomic \
+            -timeout 20m \
+            ./cmd/doco-cd/... \
+            ./internal/... \
+            ./cmd/secretproviders/awssecretmanager/... \
```
qianlongzt: change to `./cmd/secretproviders/...` ? or `./cmd/...`

cmd/secretproviders/1password/main.go
```diff
@@ -0,0 +1,46 @@
...
+	"syscall"
+
+	onepassword "github.com/kimdre/doco-cd/cmd/secretproviders/1password/internal/1password"
+	"github.com/kimdre/doco-cd/cmd/secretproviders/internal/server"
+)
+
+// Version is overridden at build time via -ldflags.
+var Version = "dev"
+
+func main() {
```
qianlongzt, line 16: Maybe we should only keep `onepassword.NewProvider` here and move all the base logic (signals, running the gRPC server) to the server package.

```go

func main() {
	if err := serve(getProvider); err != nil {
		os.Exit(1)
	}
}

type getProviderFunc func(ctx context.Context, logger *slog.Logger) (server.Provider, error)

func serve(getProvider getProviderFunc) error {
	logger := slog.New(slog.NewJSONHandler(os.Stderr, nil))
	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
	defer stop()

	endpoint := server.EndpointFromEnv()

	provider, err := getProvider(context.Background(), logger)
	if err != nil {
		logger.Error("failed to get provider", "err", err)
		return err
	}

	logger.Info("starting secret provider plugin", "provider", provider.Name(), "endpoint", endpoint)

	return server.Serve(ctx, server.Options{Endpoint: endpoint}, provider)
}

func getProvider(ctx context.Context, logger *slog.Logger) (server.Provider, error) {
	cfg, err := awssecretsmanager.GetConfig()
	if err != nil {
		return nil, fmt.Errorf("failed to load config: %w", err)
	}

	provider, err := awssecretsmanager.NewProvider(ctx, cfg.Region, cfg.AccessKeyID, cfg.SecretAccessKey)
	if err != nil {
		return nil, fmt.Errorf("failed to create provider: %w", err)
	}

	return provider, nil
}

```

.github/workflows/test.yaml
shanduur: I want to separate pure Go packages from packages requiring CGO to compile - thus the split.

shanduur: Technically after a second thought there's no reason to split that, you're right!

cmd/secretproviders/internal/server/server.go
```diff
@@ -0,0 +1,192 @@
...
+		return fmt.Errorf("failed to create listener for %q: %w", opts.Endpoint, err)
+	}
+	defer cleanup()
+
+	srv := grpc.NewServer()
+	secretproviderv1.RegisterSecretProviderServer(srv, newGRPCServer(p))
+
+	var wg sync.WaitGroup
+
+	wg.Add(1)
```
qianlongzt: use `wg.Go` ?

cmd/secretproviders/internal/server/server.go
```diff
@@ -0,0 +1,192 @@
...
+
+	l, err := lc.Listen(ctx, network, addr)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	cleanup := func() {
+		if network == "unix" {
+			_ = os.Remove(addr)
+		}
```
qianlongzt, line 131: Does TCP networking need `l.Close()`? Or does gRPC handle it?

api/secretprovider/v1/secretprovider.proto
```diff
@@ -0,0 +1,53 @@
+syntax = "proto3";
```
qianlongzt, line 1: move to `api/secretprovider/v1/` ?

I think we should keep it in the same directory.

proto/secretprovider/v1/secretprovider.proto
```diff
@@ -0,0 +1,53 @@
...
+  // GetSecrets retrieves multiple secret values in a single call.
+  rpc GetSecrets(GetSecretsRequest) returns (GetSecretsResponse);
+
+  // ResolveSecretReferences resolves a map of environment-variable names to
+  // secret references, returning the resolved values keyed by the same
+  // environment-variable names.
+  rpc ResolveSecretReferences(ResolveSecretReferencesRequest) returns (ResolveSecretReferencesResponse);
+}
+
+message NameRequest {}
```
qianlongzt: Maybe we can rename it to `Info` and return a value that includes the plugin version.
And log the version when initializing the provider.

cmd/secretproviders/internal/server/server.go
```diff
@@ -0,0 +1,192 @@
...
+	secretproviderv1 "github.com/kimdre/doco-cd/api/secretprovider/v1"
+	secrettypes "github.com/kimdre/doco-cd/internal/secretprovider/types"
+)
+
+// Provider is the in-process interface plugin binaries implement; the harness
+// adapts it to the gRPC SecretProviderServer surface.
+type Provider interface {
+	Name() string
+	GetSecret(ctx context.Context, id string) (string, error)
+	GetSecrets(ctx context.Context, ids []string) (map[string]string, error)
```
kimdre: I wonder if we even need `GetSecret()` in the public interface if we can just use `GetSecrets()`.

proto/secretprovider/v1/secretprovider.proto
kimdre: I agree, I would like to see the plugin version in the application logs.

shanduur: Will do!

cmd/secretproviders/internal/server/server.go
shanduur: gRPC server handles that internally.

api/secretprovider/v1/secretprovider.proto
shanduur: Most of the time I've seen pattern of splitting those two into separate directories, but maybe it's a good call here.

.github/workflows/test.yaml
```diff
@@ -58,6 +58,28 @@ jobs:
           version: latest
           args: --timeout=5m
 
+  build-pure-go:
```
qianlongzt: What is this for? I think `.github/workflows/build-dev-plugins.yaml` should be enough.

cmd/secretproviders/awssecretmanager/main.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+}
+
+func getProvider(ctx context.Context, _ *slog.Logger) (server.Provider, error) {
+	cfg, err := awssecretsmanager.GetConfig()
+	if err != nil {
+		return nil, fmt.Errorf("failed to load config: %w", err)
+	}
+
+	provider, err := awssecretsmanager.NewProvider(ctx, cfg.Region, cfg.AccessKeyID, cfg.SecretAccessKey)
```
qianlongzt, line 28: missing Version ?

cmd/secretproviders/bitwardensecretsmanager/main.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+}
+
+func getProvider(_ context.Context, _ *slog.Logger) (server.Provider, error) {
+	cfg, err := bitwardensecretsmanager.GetConfig()
+	if err != nil {
+		return nil, fmt.Errorf("failed to load config: %w", err)
+	}
+
+	provider, err := bitwardensecretsmanager.NewProvider(cfg.ApiUrl, cfg.IdentityUrl, cfg.AccessToken)
```
qianlongzt, line 28: +1 missing version?

cmd/secretproviders/infisical/main.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+}
+
+func getProvider(ctx context.Context, _ *slog.Logger) (server.Provider, error) {
+	cfg, err := infisical.GetConfig()
+	if err != nil {
+		return nil, fmt.Errorf("failed to load config: %w", err)
+	}
+
+	provider, err := infisical.NewProvider(ctx, cfg.SiteUrl, cfg.ClientID, cfg.ClientSecret)
```
qianlongzt, line 28: +1

.github/workflows/test.yaml
shanduur: Whoops, slipped. I'll drop it.

cmd/secretproviders/bitwardensecretsmanager/main.go
qianlongzt: Sorry, that was my mistake.

cmd/secretproviders/awssecretmanager/main.go
qianlongzt: Sorry, that was my mistake.

cmd/secretproviders/infisical/main.go
qianlongzt: Sorry, that was my mistake.

cmd/secretproviders/internal/server/server.go
qianlongzt: ok

cmd/secretproviders/awssecretmanager/main.go
shanduur: No need for that. We have the version handling directly in the `grpcServer` - that way we don't need to implement the `Info` endpoint on each of the providers, untill it is necessary.

shanduur: https://github.com/shanduur/doco-cd/blob/e2484b3a0a71fd306bde6d36dff4a13413b1e48b/cmd/secretproviders/internal/server/server.go#L165-L170

cmd/secretproviders/bitwardensecretsmanager/internal/bitwardensecretsmanager/client_nobitwarden.go
qianlongzt, line 1: Should the nobitwarden build tag be removed?

Makefile
qianlongzt, line 83: Add setup for protoc, protoc-gen-go, and buf? I suggest using [mise] to manage them.

[mise]: https://mise.jdx.dev/getting-started.html


mise.toml
```toml
[tools]
buf = "latest"
protoc = "latest"
protoc-gen-go = "latest"
protoc-gen-go-grpc = "latest"

```

qianlongzt:
Also, please avoid force pushing every time.

cmd/secretproviders/1password/main.go
```diff
@@ -0,0 +1,34 @@
...
+	if err := server.Run(Version, getProvider); err != nil {
+		os.Exit(1)
+	}
+}
+
+func getProvider(ctx context.Context, _ *slog.Logger) (server.Provider, error) {
```
qianlongzt, line 22: Maybe we should move this into the `1password` package to keep main.go smaller and cleaner.

shanduur:
@qianlongzt I am rebasing to keep the PR in sync with `main`. Each of your review changes is pushed against new `HEAD` commit, so you can easily review it.

Makefile
shanduur: I added them as a `go.mod` `tool` entries. Also, _mise_ is... ehh bad? I don't think it does anything particularly  good, and there are better options around. Even pure management of tools with Makefile is better IMHO.

qianlongzt (approved):

cmd/secretproviders/1password/main.go
shanduur: I disagree - this allows us to keep the constructor and configuration of the actual clients separate from the Server - this is a glue code that ties together `1password` package and `internal/server`. It's also relatively small already.

Makefile
qianlongzt: Mixing application dependencies and developer tooling dependencies in the same module graph is conceptually odd and can lead to unnecessary coupling and complexity.

mise is better than go mod tool; you can set versions or lock them (mise.lock), while I cannot clearly see the tool versions in go.mod.

https://news.ycombinator.com/item?id=42845323

shanduur: While the `go tool` is not perfect, it's already here in the repository. 

If I had to advise using anything else than `go tool`, I'd suggest using Nix flakes[^1] (or a derivative like devbox[^2]).

[^1]: https://wiki.nixos.org/wiki/Flakes
[^2]: https://github.com/jetify-com/devbox

kimdre: Honestly I want to keep the current Go tool approach. I don't want to manage too many different CLI tools and dependencies. Using something other than Go and Make should only be done if really necesarry.

qianlongzt: Okay, let's keep your version.

qianlongzt (approved):

qianlongzt:
Any ETA on when this will be merged? I think it will reduce the Docker image size.

kimdre:
@shanduur Did you already do some real-world tests with the provider plugins?

kimdre:
I made some changes to make the plugin discovery and CGO usage for the image builds more dynamic.

kimdre:
@shanduur I cannot test the plugin image. 

- The `SECRET_PROVIDER_GRPC_ENDPOINT` env var in the doco-cd app does not have a default value like described in the docs.
- When trying to run the plugin, the socket file never gets created.

```
secret-provider-1  | {"time":"2026-06-02T15:55:28.457616839Z","level":"ERROR","msg":"failed to construct provider","err":"failed to create provider: API error: Received error message from server: [400 Bad Request] {\"error\":\"invalid_client\"}"}
doco-cd            | {"time":"2026-06-02T17:55:28.483701513+02:00","level":"info","msg":"starting application","log_level":"info","version":"dev"}
doco-cd            | {"time":"2026-06-02T17:55:28.492404797+02:00","level":"critical","msg":"failed to initialize secret provider","error":"failed to reach secret provider plugin \"unix:///var/run/doco-cd/secret-provider.sock\": rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial unix /var/run/doco-cd/secret-provider.sock: connect: no such file or directory\""}
secret-provider-1 exited with code 1                                                                                                                                                                                                
doco-cd            | {"time":"2026-06-02T17:55:28.819204824+02:00","level":"warning","msg":"new application version available","current":"dev","latest":"v0.90.0"}
doco-cd            | {"time":"2026-06-02T17:55:28.819257043+02:00","level":"error","msg":"application stopped with error","error":"failed to reach secret provider plugin \"unix:///var/run/doco-cd/secret-provider.sock\": rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial unix /var/run/doco-cd/secret-provider.sock: connect: no such file or directory\""}
doco-cd exited with code 1 (restarting)     
```

I'm trying to execute this test compose file for bitwarden secrets manager: https://github.com/shanduur/doco-cd/blob/6a04be1ace8bb6d8add344ec8ff0ae3580720959/cmd/secretproviders/bitwardensecretsmanager/testdata/compose.yml

with the command 
```sh
docker compose -f dev.compose.yaml -f cmd/secretproviders/bitwardensecretsmanager/testdata/compose.yml up --build
```

and a .env file

```dotenv
SECRET_PROVIDER_ACCESS_TOKEN=<token>
SECRET_PROVIDER_API_URL=https://vault.bitwarden.eu/api
SECRET_PROVIDER_IDENTITY_URL=https://vault.bitwarden.eu/identity
```

shanduur:
@kimdre it fails only initially I think? The startup is a bit racy, even with the depends-on clause, I'll add a small retry. And the default value should be set from the start on the Secret Provider (https://github.com/kimdre/doco-cd/blob/73acf571197cee4daf838dcf2ceebf436ec4df87/cmd/secretproviders/internal/server/endpoint.go#L13-L27).

shanduur:
Should be good now, I don't see failure when running BitWarden plugin.

kimdre:
It still fails

```
doco-cd  | {"time":"2026-06-02T19:02:38.375110527+02:00","level":"info","msg":"starting application","log_level":"info","version":"dev"}
doco-cd  | {"time":"2026-06-02T19:02:38.712592437+02:00","level":"warning","msg":"new application version available","current":"dev","latest":"v0.90.0"}
doco-cd  | {"time":"2026-06-02T19:02:43.384966485+02:00","level":"critical","msg":"failed to initialize secret provider","error":"secret provider plugin \"unix:///var/run/doco-cd/secret-provider.sock\" socket not ready: socket \"/var/run/doco-cd/secret-provider.sock\" not found after 5s: context deadline exceeded"}
doco-cd  | {"time":"2026-06-02T19:02:43.385015384+02:00","level":"error","msg":"application stopped with error","error":"secret provider plugin \"unix:///var/run/doco-cd/secret-provider.sock\" socket not ready: socket \"/var/run/doco-cd/secret-provider.sock\" not found after 5s: context deadline exceeded"
```

When I let the secret-provider container start first, the container errors with 
```
secret-provider-1  | {"time":"2026-06-02T17:08:24.323183374Z","level":"ERROR","msg":"failed to construct provider","err":"failed to create provider: API error: Received error message from server: [400 Bad Request] {\"error\":\"invalid_client\"}"}
```

shanduur:
Secret provider must start first. Do you have more logs from both containers? I repeated what you did, and it _just works_...

kimdre:
Sorry, my fault, now it's working in my test! :) 

I have three things that should be improved in my opinion:
- [x] In the doco-cd app, the env var `SECRET_PROVIDER_GRPC_ENDPOINT` does not have a default value. It only works, when I set it manually `SECRET_PROVIDER_GRPC_ENDPOINT: unix:///var/run/doco-cd/secret-provider.sock`, otherwise doco-cd throws this error: `{"time":"2026-06-03T15:49:06.639631756+02:00","level":"error","msg":"application stopped with error","error":"failed to parse config from environment: env: environment variable \"SECRET_PROVIDER_GRPC_ENDPOINT\" should not be empty"}` 
  In the docs it says the env var uses this value by default but it's empty.
- [x] Since the secret provider logic is now located in a seperate container, it would be good to have more logs in the secret-provider container (e.g. when a new request for resolving secrets was received from a certain job) 
  - If possible, please reuse the `internal/logger` package for logging in the secret-prvoider container
  - Use the `job_id` to trace the secret resolving process between doco-cd and the secret-provider container in the logs.
- [x] Also, would it be possible to add a simple healthcheck (similiar to `cmd/doco-cd/healthcheck`) to the secret-provider container (e.g. by checking the grpc socket)?

kimdre:
@shanduur 
I found a bug with the 1Password Connect provider. It's asking for `SECRET_PROVIDER_ACCESS_TOKEN` or `SECRET_PROVIDER_ACCESS_TOKEN_FILE`, which both is only used in the standalone [1password provider](https://doco.cd/latest/External-Secrets/1Password/).
The [1Password Connect](https://doco.cd/latest/External-Secrets/1Password-Connect/) provider requires `SECRET_PROVIDER_CONNECT_TOKEN`.

Also, what do you think about renaming the `SECRET_PROVIDER` value from `grpc` to `plugin`? I would prefer that honestly. @qianlongzt What do you think?

shanduur:
Yeah, `plugin` works fine, I like it!

qianlongzt:
Personally, I'd prefer to keep `grpc` as `grpc`.

`plugin` feels too vague to me, especially since it sits alongside webhook. In my mind, "plugin" is a broader concept and could even be interpreted as including webhook-based integrations. I also tend to associate plugins more with in-process extensions, dynamically loaded modules, or direct command execution.

On a related note, I find the webhook name a bit confusing as well. Traditionally, a webhook implies that an external service calls into Doco CD. In this case, Doco CD is the one making requests to another service to retrieve secrets.

One additional concern with the `webhook` naming is that webhooks are typically event-driven callbacks where the response is just an ACK and the payload is not consumed.

In our case, Doco-CD relies on the actual response body to retrieve the secret, so it is effectively a synchronous request/response model over HTTP rather than a traditional webhook pattern.

This makes the term `webhook` somewhat misleading, and `http` would better reflect the actual behavior.

Because of that, I think keeping `grpc` as `grpc` is clearer, and I would actually lean toward renaming `webhook` to `http` instead.

shanduur:
@kimdre I've rebased the changes, please re-review.

shanduur:
Oh, looks like there were some of the commits you pushed and they are gone now @kimdre after my push. Not the best idea to work on a single PR concurrently. You might need to force push it and rebase yourselves.

PR #1377, opened 2026-05-21
docs: improve docs on encryption and nested config

JojoMee:
When I set up doco-cd in my homelab I felt that two important things were missing in the documentation, that I want to improve with this PR:
1. **How are encrypted files detected when using SOPS and which files are in scope for SOPS decryption?** I added a section to the "Encryption" page which explains this. This became quite lengthy, if you think that's too verbose/detailed I can shorten it a bit.
2. **Can I use nested config overrides when using a custom webhook target?** The answer is yes, so I added two explicit hints on the "Deploy Settings" and "Webhook Listener" pages.


wiki/docs/Advanced/Encryption.md
```diff
@@ -10,6 +10,56 @@ tags:
...
+
+- `sops`
+- `ENC[`
+
+This means you can name your encrypted files however you like. Doco-cd checks every applicable file regardless of its name or extension and skips any that don't carry these markers. No special naming convention (like `.enc.yaml`) is required.
+
+### How the file format is determined
+
+Once a file is identified as encrypted, the format used for decryption is determined **by file extension**:
+
```
kimdre, line 25: This section is too detailed in my opinion. It's probably also not necessary to explain in detail the order in which files are decrypted. Feel free to correct me if you disagree. :)

kimdre (changes requested):
Thank you for your PR! Can you please simplify your changes about the SOPS decryption process? Everything else looks good to me! 👍

wiki/docs/Advanced/Encryption.md
JojoMee: I agree, I shortend it a lot in [8fb73db](https://github.com/kimdre/doco-cd/pull/1377/commits/8fb73db2aff097b8ab4a030cdc2f3c2aafe22321), so it now mainly lists which files are in scope for decryption and briefly mentions how encrypted files are detected

kimdre:
Thank you. :)

PR #1378, opened 2026-05-22
feat(compose): add support for automatic recreation of specific named volumes

### LINKED ISSUE #1130: Add an option to indicate whether a volume should be recreated.
### Description

We also need to consider that a project may have multiple volumes—some should not be recreated, while others (e.g., tmpfs volumes) can be safely recreated.

Additionally, the auto-discovery feature makes it difficult to manage this via deploy configuration.

I think using labels is a better approach.

```
Volume "beszel-agent_socket" exists but doesn't match configuration in compose file. Recreate (data will be lost)?
```

### Use Case

```diff
name: beszel-agent
services:
  beszel-agent:
    cap_drop:
      - ALL
    container_name: beszel-agent
    image: ghcr.io/henrygd/beszel/beszel-agent:0.18.4-alpine@sha256:4855ca363653f4c481ded26765410cdd0501f7bcd05d6343d0fadab6a35e17df
    network_mode: host
    read_only: true
    restart: always
    user: 65530:65530
    volumes:
      - type: volume
        source: socket
        target: /socket
        read_only: true
        volume: {}
  socket-proxy:
    cap_drop:
      - ALL
    command:
      - -loglevel=info
      - -allowGET=/containers/json
      - -allowGET=/containers/[^/]*/json
      - -allowGET=/containers/[^/]*/stats
      - -allowGET=/containers/[^/]*/logs
      - -allowGET=/info
      - -allowGET=/version
      - -allowHEAD=/_ping
      - -proxysocketendpoint=/socket/proxy.sock
      - -proxysocketendpointfilemode=0600
      - -watchdoginterval=60
      - -stoponwatchdog
      - -shutdowngracetime=5
    image: ghcr.io/wollomatic/socket-proxy:testing-efea0f9cc41ba59fc0154e492f7118407e65e636@sha256:b66867a9db41acc1743b983524bb34fae981c53a951b02bfe98376214476d0ca
    mem_limit: "67108864"
    network_mode: none
    read_only: true
    restart: always
    user: 65530:998
    volumes:
      - type: bind
        source: /var/run/docker.sock
        target: /var/run/docker.sock
        bind: {}
      - type: volume
        source: socket
        target: /socket
        volume: {}
volumes:
  socket:
    name: beszel-agent_socket
    driver: local
    driver_opts:
      device: tmpfs
      o: size=1k,uid=65530,gid=0,mode=0700,noexec
      type: tmpfs
+    labels:
+      "cd.doco.volume.recreate": true
```

### Additional Information

_No response_

kimdre:
### Allow recreation of specific named volumes

When Docker Compose detects a named volume definition change, deployment can fail with:

```text
Volume "<name>" exists but doesn't match configuration in compose file. Recreate (data will be lost)?
```

To opt in to automatic recreation for volumes that are safe to recreate (for example tmpfs-backed volumes),
set the top-level volume label `cd.doco.volume.recreate: "true"`.

Only volumes with this label are eligible. Unlabeled volumes are never removed automatically.

```yaml
volumes:
  socket:
    name: beszel-agent_socket
    driver: local
    driver_opts:
      type: tmpfs
      device: tmpfs
      o: size=1k,uid=65530,gid=0,mode=0700,noexec
    labels:
      cd.doco.volume.recreate: "true"
```

### DIFF
--- a/internal/docker/compose.go
+++ b/internal/docker/compose.go
@@ -11,6 +11,7 @@ import (
 	"path"
 	"path/filepath"
 	"reflect"
+	"regexp"
 	"slices"
 	"strconv"
 	"strings"
@@ -52,6 +53,7 @@ var (
 	ErrNoContainerToStart = errors.New("no container to start")
 	ErrIsInUse            = errors.New("is in use")
 	ComposeVersion        string // Version of the docker compose module, will be set at runtime
+	volumeMismatchRegex   = regexp.MustCompile(`Volume "([^"]+)" exists but doesn't match configuration in compose file`)
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


qianlongzt:
I’m not sure this should be handled in doco-cd. Maybe it belongs in a fork or should be upstreamed to `docker/compose`.

I opened a PR to Docker Compose. https://github.com/docker/compose/pull/13808

PR #1427, opened 2026-06-09
feat(hooks): per-deployment HTTP webhook hooks

foofybuster:
## What

Adds per-deployment HTTP webhook hooks. A deploy config can declare `hooks.on_success` and `hooks.on_failure` lists; each entry POSTs a JSON payload to an endpoint when the deployment succeeds or fails.

## Why

Lets external systems (CI, alerting, chatops) react to doco-cd deployments without polling — e.g. notify a channel on success or page on failure.

## Config

```yaml
name: some-project
hooks:
  on_success:
    - url: https://ci.example.com/notify
      headers:
        Authorization: Bearer xxx
  on_failure:
    - url: https://alerts.example.com/webhook
      method: PUT   # defaults to POST
```

Fields per entry: `url` (required, http/https), `method` (default `POST`), `headers` (optional map).

## Payload

JSON body (`Content-Type: application/json`):

| Field | Description |
|-------|-------------|
| `event` | `success` or `failure` |
| `repository` | source repo / OCI artifact name |
| `stack` | deployment / stack name |
| `revision` | deployed reference + commit (e.g. `main (abc1234)`) |
| `job_id` | unique job id |
| `images` | resolved image refs (`name:tag`) of the **changed** services; omitted when empty |
| `error` | failure reason (failure events only) |

## Behaviour

- Success hooks fire from the post-deploy stage; failure hooks from `StageManager.NotifyFailure`.
- **Best-effort**: 10s timeout per request; a non-2xx response or unreachable endpoint is logged but never fails the deployment.
- Config validated up front (`Hooks.Validate` rejects empty / non-http(s) URLs).
- Multi-doc deploy YAML and `allowOverride` per-target overrides supported.

## Tests

- `internal/hook`: URL validation, send (method/headers/body), non-2xx + bad-URL paths.
- `internal/config/deploy`: hook parsing from YAML, invalid-URL rejection, multi-doc handling.
- `internal/stages`: `changedServiceImages` (dedup/sort, missing-image, nil-project).

## Notes

`images` currently reflects only **changed** services, so it is omitted on deploys with no file-detected service changes (e.g. force-recreate / mismatch-driven redeploys). Can be widened to all stack services if desired.

### DIFF
--- a/internal/config/deploy/deploy.go
+++ b/internal/config/deploy/deploy.go
@@ -21,6 +21,7 @@ import (
 	"gopkg.in/validator.v2"
 
 	"github.com/kimdre/doco-cd/internal/config"
+	"github.com/kimdre/doco-cd/internal/hook"
 
 	secrettypes "github.com/kimdre/doco-cd/internal/secretprovider/types"
 
@@ -66,6 +67,7 @@ type Config struct {
 	AutoDiscovery      AutoDiscoveryConfig                      `yaml:"auto_discovery" json:"auto_discovery"`                                                                                                                   // AutoDiscovery configures autodiscovery of services to deploy in the working directory
 	Reconciliation     ReconciliationConfig                     `yaml:"reconciliation" json:"reconciliation" doco:"allowOverride"`                                                                                              // Reconciliation is the configuration for the reconciliation feature
 	Oci                config.OciTrustPolicyOverride            `yaml:"oci" json:"oci" doco:"allowOverride"`                                                                                                                    // Oci allows per-target overrides for OCI signature verification policy
+	Hooks              hook.Config                              `yaml:"hooks" json:"hooks" doco:"allowOverride"`                                                                                                                // Hooks configures HTTP webhook hooks fired on deployment success/failure
 	Internal           struct {
 		File                          string            `yaml:"-"` // File is the path to the deployment configuration file
 		Environment                   map[string]string // Environment stores environment variables for variable interpolation in the compose project
@@ -182,6 +184,10 @@ func (c *Config) Validate() error {
 
 	if err := c.normalizeReconciliationEvents(); err != nil {
 		return err
+	}
+
+	if err := c.Hooks.Validate(); err != nil {
+		return fmt.Errorf("%w: %v", ErrInvalidConfig, err)
 	}
 
 	return nil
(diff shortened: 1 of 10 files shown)


kimdre:
Does this also work with auto-discovery and remote repositories?

foofybuster:
> Does this also work with auto-discovery and remote repositories?

I've already tested on remote repositories in my environments. 

Let me validate whether it works properly with auto-discovery.

foofybuster:
@kimdre I'm sorry for taking too long.
I have a concern about this PR. I changed some behavior of deployment settings when `force_image_pull: true` and `force_recreate: true` 

Should Doco-cd force to pull an image or recreate a container when the Doco-cd is restarted? [e.g. `docker restart doco-cd`]

As I understood your current version will not force when the Doco-cd only restarts. Is it a bug or your intent? 
Note that if there are any change in a service's `working_dir` and then restarting Doco-cd cause recreating the service's container.

kimdre:
> Should Doco-cd force to pull an image or recreate a container when the Doco-cd is restarted? [e.g. `docker restart doco-cd`]
>
> As I understood your current version will not force when the Doco-cd only restarts. Is it a bug or your intent? Note that if there are any change in a service's `working_dir` and then restarting Doco-cd cause recreating the service's container.

Doco-CD is designed to mirror Docker's default behavior; Image pulls and container recreations should only happen on deployment/compose spec/config drift or when explicity configured.

foofybuster:
> > Should Doco-cd force to pull an image or recreate a container when the Doco-cd is restarted? [e.g. `docker restart doco-cd`]
> > As I understood your current version will not force when the Doco-cd only restarts. Is it a bug or your intent? Note that if there are any change in a service's `working_dir` and then restarting Doco-cd cause recreating the service's container.
> 
> Doco-CD is designed to mirror Docker's default behavior; Image pulls and container recreations should only happen on deployment/compose spec/config drift or when explicity configured.

I got it. I will rollback to the designed behavior.

qianlongzt:
Maybe we can change the webhook payload to use a Golang template or something similar, allowing users to customize it.


For example, CrowdSec notifications can use Golang templates.
https://docs.crowdsec.net/docs/local_api/notification_plugins/http

foofybuster:
> Maybe we can change the webhook payload to use a Golang template or something similar, allowing users to customize it.
> 
> For example, CrowdSec notifications can use Golang templates. https://docs.crowdsec.net/docs/local_api/notification_plugins/http

Noted. So I close this PR first. I will open another PR soon. Thank you for your idea.

PR #1558, opened 2026-07-26
feat(notification): add configurable Go template for notification body

dennislapchenko:
## What

Adds an optional `APPRISE_NOTIFY_TEMPLATE` (+ `APPRISE_NOTIFY_TEMPLATE_FILE`) environment variable holding a [Go `text/template`](https://pkg.go.dev/text/template) that renders the notification **body**. When unset, the built-in format is used **unchanged** — fully backward compatible.

## Why

Today the body is hard-coded in `internal/notification/notification.go`: the message followed by alphabetically-sorted `key: value` metadata lines. There is no way to drop noisy fields (e.g. `job_id`), add a host label, or collapse to a one-liner without standing up a custom HTTP endpoint in front of the Apprise sidecar. This is painful when several stacks/hosts report into one channel.

## How

- The template lives with the other **global** Apprise settings (URL, notify URLs, level) and is applied in `Send()` for every notification. This keeps configuration consistent and covers call sites that have no deploy config in scope (startup ping, poll-job failure), which a per-`.doco-cd.yml` field could not.
- Parsed **and executed against a sample** at startup with `missingkey=error`, so a syntax error or an unknown field reference stops doco-cd from starting instead of silently dropping alerts.
- On the rare runtime execution error, the body **falls back** to the built-in format so an alert is never lost.
- The **title** (emoji + optional `[R]` marker + text) is intentionally left as-is.

### Fields exposed to the template

`.Level`, `.Emoji`, `.Title`, `.Message`, `.IsReconciliation`, plus all metadata: `.Repository`, `.Stack`, `.Revision`, `.JobID`, `.ReconciliationEvent`, `.TraceID`, `.AffectedActorKind`, `.AffectedActorID`, `.AffectedActorName`.

### Example

```yaml
environment:
  APPRISE_NOTIFY_TEMPLATE: "{{.Emoji}} {{.Stack}} — {{.Message}} ({{.Revision}})"
```

renders e.g. `✅ app — Successfully deployed stack app (main (abc123))`.

## Tests / docs

- Unit tests for template parse validation (syntax error + unknown field rejected at config time) and rendering (metadata fields, reconciliation flag).
- Existing `SetAppriseConfig` signature updated (one extra arg); its one non-test caller and the test are updated.
- Docs updated in `wiki/docs/Advanced/Notifications.md` (settings table + new "Custom notification body" section).
- `gofmt` clean, `golangci-lint run` reports 0 issues on the changed packages, `go build ./...` passes.

## Notes

- Body-only by design (title semantics preserved). Happy to extend to the title too if you'd prefer.
- Open to making this a per-`.doco-cd.yml` field instead of / in addition to the global env var if that fits the roadmap better — went global for consistency with the existing Apprise config and to cover non-deploy notifications.


kimdre:
Thanks for the PR! :)

Two things:
- Could you please rename the `parseTemplate` function and its corresponding test to `validateTemplate`?
- It would probably also make sense to replace the old `formatMessage` function with a default template.

dennislapchenko:
Thanks for the review! Both addressed in 2d677b9:

- Renamed `parseTemplate` → `validateTemplate` (and `TestParseTemplate` -> `TestValidateTemplate`).
- Replaced `formatMessage` with a default template. The built-in body logic now lives in `TemplateData.DefaultBody`, adn `defaultTemplate` (`{{ .DefaultBody }}`) is used whenever `APPRISE_NOTIFY_TEMPLATE` is unset. `Send` always renders through a template now (no branch). As a bonus , `.DefaultBody` is exposed to user templates so they can extend the built-in format instead of replacing it, e.g. `{{ .DefaultBody }}\nhost: my-vm`.

kimdre:
I'm testing your changes right now and it seems to work flawlessly. :+1: 

1. Can you please rename `APPRISE_NOTIFY_TEMPLATE` to `APPRISE_NOTIFY_BODY_TEMPLATE`. I think this would make it more clear to users what it does.
2. This is optional, I'm not sure how much work it would be, but could you maybe add the Docker Context (introduced in [v0.100.0](https://github.com/kimdre/doco-cd/releases/tag/v0.100.0)) to the available template fields as `.Context`?

dennislapchenko:
Both done.
 - Renamed to `APPRISE_NOTIFY_BODY_TEMPLATE`.
 - Added the Docker context as `.Context` — bit more involved - but it's populated for deploy/destroy and reconciliation-restart notifications from the deploy config. Left it out of the default body so the built-in format stays the same, but it's available for custom templates now.

kimdre (approved):

kimdre:
I found a bug in the version notification, when the body template variable is set to something like
```
{{.Emoji}} {{.Stack}} ({{if .Context}}{{.Context}}{{else}}default{{end}}) — {{.Message}} ({{.Revision}})
```

The message shows an emoji, the context and the empty bracket of the empty revision:

```
ℹ️ New version of doco-cd is available

ℹ️  (default) — Current Version: dev
Latest Version: v0.102.0

https://github.com/kimdre/doco-cd/releases ()
```

but it should be

```
ℹ️ New version of doco-cd is available

Current Version: dev
Latest Version: v0.102.0

https://github.com/kimdre/doco-cd/releases
```

dennislapchenko:
Good catch, thanks. I think it is not really bug in `.Context` itself, problem is a bit different.

Version notification (`cmd/doco-cd/version.go`) is sent with empty `Metadata{}` — no stack, no context, no revision. It is app-level "new version available" ping, not deployment. But body template is written for deploy notification, so when all these fields are empty, you get what you saw. Both strange things come from the same reason:

- `(default)` is your own `{{if .Context}}{{else}}default{{end}}` doing what you asked: prints "default" when context is empty. For real deploy on default Docker context this is even correct — it only looks wrong here because version ping should not go through deploy template at all.
- ` ()` are literal brackets from the template (`({{.Revision}})`) with empty `.Revision`, rendered also on notification which has no revision.

So the output you want is basically just the message, which is exactly what default body already produces for empty metadata.

One more thing: version ping is not the only notification with thin metadata. "Poll Job failed" has no stack, and "Deployment Failed" from webhook has no stack and no revision, so the same template prints empty stack / `()` there too. Probably you did not hit them because version ping fires on every start while testing.

So two ways to fix:

A) Only version notification skips the template and falls back to default body (= plain message, the output you want). Deploy and failure notifications stay templated — author guards optional fields with `{{if}}`, same like you already did for context. Smallest change, does not take control away from the template.

B) Skip template for any notification without a stack. Fixes all three at once, but then even a template that guards with `{{if .Stack}}` is ignored on failure pings, and maybe someone wants them.

On my side I would go with A — cleaner and matches your expected output one to one. But your call, tell me which you prefer and I do it.

kimdre:
A is fine for me

dennislapchenko:
Implemented. I thought going with options pattern is best for clean exec of it. lmk what you think

kimdre (approved):

kimdre:
Perfect, thank you! :)

PR #1564, opened 2026-07-27
fix(git): allow custom HTTP auth username for token clone

dennislapchenko:
Token clone over HTTP(S) always used `oauth2` as the basic auth username. For GitLab deploy tokens this doesn't work — username must be exactly the token name (e.g. `gitlab+deploy-token-123`), so the only workaround was putting `user:token@` into the repo URL. That then leaks the token into logs and notifications when something fails (like a fetch error).

Now the username is configurable:

- `GIT_HTTP_AUTH_USER` — global, paired with `GIT_ACCESS_TOKEN`
- `http_auth_user` — per entry in `GIT_AUTH_DOMAINS`

Empty stays `oauth2`, so nothing changes for GitHub and the others.

`AUTH_TYPE` was already documented for exactly this purpose but was never wired to anything (dead config). Instead of adding a second knob i renamed it to `GIT_HTTP_AUTH_USER`. Since it never did anything, nobody could depend on old behaviour, so no real break.

Docs and `GIT_AUTH_DOMAINS` example updated, tests cover default fallback plus custom username through both global and scoped path.


wiki/docs/Git-Settings.md
```diff
@@ -32,7 +32,7 @@ You can use either
...
-| `AUTH_TYPE`                       | string | AuthType is the type of authentication to use when cloning repositories via **http**.                                                                                                                                                                                                   | `oauth2`                                         |
+| `GIT_HTTP_AUTH_USER`              | string | HTTP Basic Auth **username** paired with `GIT_ACCESS_TOKEN` when cloning via **HTTP(S)**. Most providers accept any non-empty value, but GitLab **deploy tokens** require their specific username (e.g. `gitlab+deploy-token-1234567`). Setting this avoids embedding credentials in the repository URL.                                                                                                                                                                                                   | `oauth2`                                         |
```
kimdre: Please remove the text `Setting this avoids embedding credentials in the repository URL.` 
I don't think it's useful there.

kimdre: The naming of `GIT_HTTP_AUTH_USER` and `http_auth_user` is too generic for me. I would prefer something more explicit and unambiguous like `git_access_token_user` / `GIT_ACCESS_TOKEN_USER` so users know exactly what it does and that it pairs with `git_access_token` / `GIT_ACCESS_TOKEN` when they read their configs.

wiki/docs/Git-Settings.md
```diff
@@ -81,6 +82,7 @@ Each entry in the list has the following structure:
 |------------------------------|--------|----------|----------------------------------------------------------------------------------------------------------------------|
 | `domains`                    | list   | Yes      | List of domain names to apply these credentials to. Supports exact domains and wildcard patterns.                    |
 | `git_access_token`           | string | No       | HTTP(S) access token for authenticating with the Git provider. Cannot be used with `ssh_private_key`.                |
+| `http_auth_user`             | string | No       | HTTP Basic Auth username paired with `git_access_token`. Defaults to `oauth2`. Set to the deploy-token username for GitLab deploy tokens (e.g. `gitlab+deploy-token-1234567`). |
```
kimdre: Can you check if setting a username is also required when using deploy tokens from other providers like Github or Gitea?

kimdre (changes requested):
Thanks for the PR :) I commented some things that should be improved.

wiki/docs/Git-Settings.md
dennislapchenko: I am up for the renames, but got no time bandwidth to go testing other providers.
- We can either make comment provider agnostic, simply saying that if provider needs the user as well this is what you'd use
- you can test yourself if you'd like
- or you can reject the whole thing if you'd like of course 😄

kimdre: I think it's fine if we documented this provider agnostic like you said. It makes sense to specify the deploy token user in any case since it makes it also clear which deploy token is used.

dennislapchenko: all done. left gitlab deploy token as an example in brackets

kimdre (approved):

kimdre:
Thank you for your work

PR #1582, opened 2026-07-29
feat(notification): expose deployment target in notification templates

JojoMee:
I'm using [custom webhook targets](https://doco.cd/latest/Endpoints/Webhook-Listener/#with-custom-target) to deploy to multiple servers from a single repository. In my notifications I would like to have a reference to the deployment target, but this value was missing from the notification metadata added in #1558. I added it to the metadata and you can now use it in custom notification body templates:
```
{{.Message}}

{{if .Target}}target: {{.Target}}
{{end}}job_id: {{.JobID}}
revision: {{.Revision}}
stack: {{.Stack}}
```
I tested this successfully with ntfy:
image
image

The implementation was done using GitHub Copilot with GPT-5.5 but I reviewed it manually and tested it successfully on my server.

### DIFF
--- a/cmd/doco-cd/handler.go
+++ b/cmd/doco-cd/handler.go
@@ -192,6 +192,7 @@ func handle(ctx context.Context, jobLog *slog.Logger,
 
 	if customTarget != "" {
 		jobLog = jobLog.With(slog.String("target", customTarget))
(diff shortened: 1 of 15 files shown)


kimdre (approved):
looks good to me :)

kimdre:
Thanks for the PR!

PR #1588, opened 2026-07-30
feat(notification): expose deployed commits in body templates

dennislapchenko:
Notification body templates could show only the single `{{ .Revision }}` string. This adds `{{ .Commits }}` - the list of commits deployed since the previously deployed commit - so a new pull can render a changelog in the notification.

Each entry exposes `.ShortHash`, `.Subject`, `.Author`. Printing an entry directly (`{{ . }}`) gives `shortHash subject`. Example:

```yaml
environment:
  APPRISE_NOTIFY_BODY_TEMPLATE: |
    {{ .DefaultBody }}
    {{ range .Commits }}- {{ .ShortHash }} {{ .Subject }} ({{ .Author }})
    {{ end }}
```

How it works:

- Previously deployed SHA is read from the container label in pre-deploy (stage_2) and carried on `DeploymentState` to `post-deploy` (stage_4).
- `git.GetCommitsBetween` walks the log from the new commit back to the merge-base with the old one, so after a rebase or force-push only the diverged commits are listed, not the whole branch. Ordered by committer time.
- `Metadata.Commits` is the new template field. The default body is untouched - nothing shows unless a template ranges over `.Commits`.

Edge cases:

- First deploy, OCI sources, or a redeploy with no new commits give an empty list.
- Capped at 50 commits.
- Best-effort: any walk error is logged and the notification still sends with an empty list.
- Shallow clones (`GIT_CLONE_DEPTH > 0`): if the old commit is beyond the shallow boundary the list may be truncated or empty. Documented in the notifications page.

Tests in internal/git/commits_test.go cover ordering, empty when old==new, the cap, and a diverged (force-push) history stopping at the merge-base.

### DIFF
--- a/internal/git/commits_test.go
+++ b/internal/git/commits_test.go
@@ -18,7 +18,7 @@ func commitN(t *testing.T, wt *gogit.Worktree, n int) []plumbing.Hash {
 	hashes := make([]plumbing.Hash, 0, n)
 	when := time.Date(2026, 1, 1, 0, 0, 0, 0, time.UTC)
 
-	for i := 0; i < n; i++ {
+	for i := range n {
 		sig := &object.Signature{Name: "Jane Doe", Email: "jane@example.com", When: when.Add(time.Duration(i) * time.Minute)}
 
 		h, err := wt.Commit("commit "+string(rune('a'+i)), &gogit.CommitOptions{
--- a/internal/git/commits_test.go
+++ b/internal/git/commits_test.go
@@ -1,0 +1,126 @@
+package git
+
+import (
+	"testing"
+	"time"
+
+	"github.com/go-git/go-billy/v5/memfs"
+	gogit "github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/storage/memory"
+)
+
+// commitN creates n empty commits and returns their hashes oldest-first.
+func commitN(t *testing.T, wt *gogit.Worktree, n int) []plumbing.Hash {
+	t.Helper()
+
+	hashes := make([]plumbing.Hash, 0, n)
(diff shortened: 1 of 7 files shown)


internal/notification/notification.go
```diff
@@ -110,6 +113,9 @@ func validateTemplate(tmpl string) (*template.Template, error) {
 		Metadata: Metadata{
 			Repository: "github.com/acme/app", Stack: "app", Context: "default",
 			Revision: "refs/heads/main (abc123)", JobID: "sample",
+			Commits: []git.CommitInfo{
+				{Hash: "abc123", ShortHash: "abc123", Subject: "sample commit", Author: "Jane Doe"},
```
kimdre, line 117: `git.CommitInfo` also provides a `Hash` value to the template, but this is not mentioned in the documentation.

internal/git/commits_test.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+// commitN creates n empty commits and returns their hashes oldest-first.
+func commitN(t *testing.T, wt *gogit.Worktree, n int) []plumbing.Hash {
+	t.Helper()
+
+	hashes := make([]plumbing.Hash, 0, n)
+	when := time.Date(2026, 1, 1, 0, 0, 0, 0, time.UTC)
+
+	for i := 0; i < n; i++ {
```
kimdre:
```go
for i := 0; i < n; i++ {
```

can be replaced with

```go
for i := range n {
```

kimdre (changes requested):
Thanks for the PR. Please see my comments for changes.

internal/git/commits_test.go
dennislapchenko: fixed

internal/notification/notification.go
dennislapchenko: fixed

kimdre (approved):
Thanks :)

PR #1591, opened 2026-07-31
feat(notification): expose deploy duration and changed services in body templates

dennislapchenko:
Two more template fields for the notification body, following the same pattern as `{{ .Commits }}`

`{{ .Duration }}` - time from job start to the notification. Useful to spot a slowly creeping deploy time right in the ping. Set on deploy, destroy and failure notifications; stays zero where no deployment ran (reconciliation restarts, scheduled jobs), so `{{ if .Duration }}` hides it there. Truncated to milliseconds, renders via `time.Duration.String()` (`12.483s`).

`{{ .ChangedServices }}` - sorted unique names of the services this deploy force-recreated (changed mounted/referenced files or a changed auto-discovery label). Answers exactly which service in the stack was deployed. Empty when the whole stack is (re)deployed for other reasons - compose config change, image update, drift, `force_recreate`.

Example:

```yaml
environment:
  APPRISE_NOTIFY_BODY_TEMPLATE: |
    {{ .Stack }} @ {{ .Revision }}{{ if .Duration }} in {{ .Duration }}{{ end }}
    {{- range .ChangedServices }}
    recreated: {{ . }}
    {{- end }}
```

How it works:

- `Metadata.Duration` / `Metadata.ChangedServices` are the new fields; both are in the startup validation sample, so templates referencing them fail fast on config errors as before. Default body is untouched.
- Duration is `time.Since(Stages.Init.StartedAt)` in post-deploy/post-destroy; in `NotifyFailure` it is guarded by `IsZero` for failures that happen before init.
- Changed services flatten `DeploymentState.changedServices` (the list stage_3 force-recreates), deduped and sorted for stable output - `slice.Unique` is map-backed, order was random otherwise.
- Hand-built metadata paths (reconciliation restarts, scheduler, version ping) simply leave both at zero values.

Edge cases: first deploy and OCI sources give an empty `ChangedServices`; a drift-only or config-only redeploy too (nothing was file-triggered). Tests cover template render of both fields plus the flatten/dedupe/sort helper.

Docs row added to the notifications page.

### DIFF
--- a/internal/notification/notification.go
+++ b/internal/notification/notification.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"sync"
 	"text/template"
+	"time"
 
 	"github.com/kimdre/doco-cd/internal/git"
 )
@@ -79,6 +80,8 @@ type Metadata struct {
 	AffectedActorID     string
 	AffectedActorName   string
 	Commits             []git.CommitInfo // commits deployed since the last deploy; empty on first deploy/failure/OCI
+	Duration            time.Duration    // time from job start to the notification; zero when no deploy/destroy ran
+	ChangedServices     []string         // services force-recreated by this deploy; empty when the whole stack is (re)deployed
 }
 
 // TemplateData is the data exposed to a user-configured notification body template.
@@ -116,6 +119,8 @@ func validateTemplate(tmpl string) (*template.Template, error) {
 			Commits: []git.CommitInfo{
 				{Hash: "abc123", ShortHash: "abc123", Subject: "sample commit", Author: "Jane Doe"},
 			},
+			Duration:        42 * time.Second,
+			ChangedServices: []string{"app"},
 		},
 	}
 	if err := t.Execute(io.Discard, sample); err != nil {
--- a/internal/notification/notification_test.go
+++ b/internal/notification/notification_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/kimdre/doco-cd/internal/common/id"
 	"github.com/kimdre/doco-cd/internal/test"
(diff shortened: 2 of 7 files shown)


dennislapchenko:
@kimdre while dogfooding the body template we hit one thing worth deciding on: a template that doesn't reference `.Message` silently swallows the failure reason. The title only says `❌ Deployment failed`, and on success `.Message` is redundant with the title — so it's very natural to write an identity one-liner and only later discover that failure pings carry no error at all (we did exactly that).

Ideally the error should probably survive regardless of the template. Two options we see:

1. hard guarantee: on `failure` level append `.Message` to the rendered body when the template output doesn't already contain it
2. soft: document the pitfall in the wiki and recommend `{{ if ne .Level "success" }} — {{ .Message }}{{ end }}` as part of every custom template example

1 is a bit magical but a safer default, 2 keeps rendering fully predictable. Which way do you prefer? Happy to send a follow-up PR for either.

dennislapchenko:
@kimdre one more thought, separate from this PR: perhaps some of the recent template-only fields deserve a place in the default body too. `duration: 12.483s` fits naturally as one more metadata line, and the commit changelog from #1588 is arguably useful for every user, not only for template authors. It's a taste question though, and it changes output for everyone with default settings — so leaving it entirely to you, outside of this PR, if you want it at all.

kimdre:
> 1 is a bit magical but a safer default, 2 keeps rendering fully predictable. Which way do you prefer? Happy to send a follow-up PR for either.

I would prefer 2. It's documented behaviour and users can always check in the logs if they want to know what's the problem.

kimdre:
> @kimdre one more thought, separate from this PR: perhaps some of the recent template-only fields deserve a place in the default body too. `duration: 12.483s` fits naturally as one more metadata line, and the commit changelog from #1588 is arguably useful for every user, not only for template authors. It's a taste question though, and it changes output for everyone with default settings — so leaving it entirely to you, outside of this PR, if you want it at all.

Feel free to add the duration to the default body metadata. I would however not add the commits by default because it can make the body very long.

kimdre (approved):

PR #1594, opened 2026-08-01
feat(notification): duration in default body, document the .Message pitfall

dennislapchenko:
Follow-ups from the #1591 discussion:

- Default body gets a `duration: 12.483s` metadata line. Only when a deployment actually ran: zero duration is omitted, so reconciliation restarts, scheduled jobs and the version ping keep their current bodies byte-for-byte. 

- Updated docs with `.Message` ting
```
{{ if ne .Level "success" }} — {{ .Message }}{{ end }}
```

Test covers the new line and the omitted-when-zero case; existing DefaultBody expectations untouched.

### DIFF
--- a/internal/notification/notification.go
+++ b/internal/notification/notification.go
@@ -309,6 +309,10 @@ func (d TemplateData) DefaultBody() string {
 		fields["job_id"] = m.JobID
 	}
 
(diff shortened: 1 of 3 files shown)


kimdre:
Thanks :)

PR #1626, opened 2026-08-05
docs: document the Docker API permissions required for each feature

### LINKED ISSUE #1624: Document the minimum Docker API permissions required for each doco-cd feature
## Description

I am using doco-cd `v0.106.0` through a restricted Docker socket proxy rather than mounting the Docker socket directly.

The documentation explains how to connect doco-cd to Docker, but it does not provide a least-privilege list of Docker API endpoints and HTTP methods required for the different doco-cd features.

As a result, the required permissions currently have to be discovered through failed deployments and proxy logs.

For example, I encountered `403 Forbidden` responses when doco-cd attempted operations such as:

```text
POST /configs/<id>/update
POST /networks/create
```

Allowing every Docker API endpoint makes the deployment work, but removes most of the security benefit of using a socket proxy.

## Requested documentation

Please document the Docker API permissions required for each operating mode and optional feature.

A table similar to the following would be useful:

| Feature | Docker resources | Access type |
| --- | --- | --- |
| Startup and discovery | info, version, containers, services, tasks, nodes, networks, configs, secrets | read |
| Compose deployment | containers, images, networks, volumes | read/write |
| Swarm stack deployment | services, tasks, networks, configs, secrets | read/write |
| Config rotation | configs and service updates | read/write |
| Secret rotation | secrets and service updates | read/write |
| Image pulling | images and registry authentication | read/write |
| Image pruning | images | destructive |
| Orphan removal | services, containers, networks, or volumes | destructive |
| Stack destruction | services, networks, configs, and secrets | destructive |
| Reconciliation | relevant discovery and update endpoints | read/write |

The table above is only an example; the documentation should contain the actual endpoints and methods used by doco-cd.

Ideally, it would identify:

- endpoints required in all installations;
- endpoints required only for Docker Compose;
- endpoints required only for Docker Swarm;
- permissions enabled by optional settings such as `prune_images`, `remove_orphans`, `destroy`, and reconciliation;
- destructive permissions that can remain disabled when the corresponding features are disabled.

## Example socket-proxy policy

It would also help to provide a tested example for a supported Docker socket proxy.

The example should show a conservative baseline for normal deployments and separate opt-in permissions for destructive features. It should clearly state that proxy implementations may use different environment-variable names and that the Docker API method and endpoint list is the source of truth.

## Expected outcome

An operator should be able to configure doco-cd with only the Docker API permissions needed for their selected features, without repeatedly broadening access after runtime `403` errors.

This would improve both security and usability for production deployments that intentionally avoid exposing the unrestricted Docker socket.

nosovk:
## Description

Adds a documentation page listing the Docker Engine API endpoints doco-cd uses, grouped by feature, so operators running doco-cd behind a Docker socket proxy can configure least-privilege access instead of discovering the required permissions through failed deployments and proxy logs.

Closes #1624

## How the endpoint list was produced

Rather than deriving the list from the source alone, I ran a doco-cd instance with its Docker socket replaced by a transparent logging proxy and recorded every API request made during real deployments:

| Scenario | Result |
|---|---|
| Startup / swarm detection | captured |
| Compose deployment | 3 containers deployed |
| Compose destroy | stack removed |
| REST API (`restart`, `stop`, `start`) | captured |
| Swarm stack deployment | services + configs + secrets + overlay networks |
| Swarm redeployment | config/secret/service updates |
| Swarm stack destroy | stack removed |

The observed calls were then cross-checked against a static analysis of the call sites in `internal/docker/`, `internal/docker/swarm/`, `internal/reconciliation/`, `internal/scheduler/` and the `docker/compose/v5` operations doco-cd invokes.

## Answers to the specific questions in the issue

**`POST /networks/create`** comes from `internal/docker/swarm/deploy_composefile.go:231` (`createNetworks`), which defaults the driver to `overlay`.

**`POST /configs/{id}/update`** comes from `internal/docker/swarm/deploy_composefile.go:188` (`createConfigs`). It is only reached when redeploying an **unchanged** config: doco-cd appends a content hash to config names, so a new or changed config is created via `POST /configs/create`, while an unchanged one already exists under its hashed name and is refreshed via update. Both endpoints are therefore required. The same applies to secrets.

## Findings worth highlighting

The example table in the issue is a good starting point, but a few endpoints are easy to miss:

- **`GET /nodes`** is required in Swarm mode. doco-cd waits for services to converge, which reads the list of active nodes. Most socket proxies gate this behind a separate `NODES` permission that is **off by default**, which makes every Swarm deployment fail with `403 Forbidden`.
- **`GET /events`** is needed by default, since both reconciliation and the scheduler are enabled out of the box.
- **`GET /distribution/{name}/json`** is used for image update detection.
- **`prune_images` in Swarm mode bypasses the proxy entirely.** It runs `docker image prune` inside a global job service that bind-mounts `/var/run/docker.sock` on every node (`internal/docker/swarm_job.go`), so the raw socket must be available on the workers. This is documented as a caveat.

The page also states plainly that a socket proxy limits the blast radius but cannot make access read-only: doco-cd is a deployment tool and needs write access to containers, images, networks and volumes (plus services, configs and secrets in Swarm mode). In practice only `/build`, `/exec`, `/auth`, `/plugins`, `/session` and `/system/df` can be denied outright, and those are listed explicitly.

## Verification of the example configuration

The example was tested end to end against `tecnativa/docker-socket-proxy` on a single-node Swarm:

- With the documented permissions, a full Swarm deployment (services, configs, secrets, overlay networks) completes with **0 × 403**.
- Removing `CONFIGS`, `SECRETS` and `NODES` reproduces exactly the failure reported in the issue:

  ```text
  failed to deploy swarm stack probe-stack: Error response from daemon:
  <html><body><h1>403 Forbidden</h1>
  Request forbidden by administrative rules.
  </body></html>
  ```

A troubleshooting section maps the most common missing permissions to their symptoms.

## Changes

- `wiki/docs/Advanced/Docker-API-Permissions.md` — new page
- `wiki/zensical.toml` — nav entry under **Advanced**
- Cross-links added from `Docker-Settings.md`, `Advanced/Swarm-Mode.md` and `Getting-Started.md`

## Checks

- `zensical build` — `No issues found` (all internal links and anchors resolve)
- `codespell` with the CI settings — clean
- Documentation only; no code changes

## Note

My commit is unsigned, contrary to the guidance in `CONTRIBUTING.md` — I do not have a signing key set up on this machine. Happy to re-sign or squash if you would prefer that before merging.

### DIFF
--- a/wiki/docs/Advanced/Docker-API-Permissions.md
+++ b/wiki/docs/Advanced/Docker-API-Permissions.md
@@ -93,9 +93,20 @@ These endpoints are only needed when the corresponding setting is enabled.
 | [`destroy.remove_volumes`](../Deploy-Settings.md)                     | `true`  | `GET /volumes/{name}`, `DELETE /volumes/{name}`                                         | destructive |
 | [`destroy.remove_images`](../Deploy-Settings.md)                      | `true`  | `DELETE /images/{name}`                                                                 | destructive |
 | [`reconciliation.enabled`](../Deploy-Settings.md)                     | `true`  | `GET /events`, `POST /containers/{id}/restart`                                          | read/write  |
-| `SCHEDULER_ENABLED`                                                   | `true`  | `POST /containers/create`, `POST /containers/{id}/wait`, `POST /containers/{id}/start`, Swarm: `POST /services/create`, `DELETE /services/{id}`, `GET /tasks` | read/write  |
+| `SCHEDULER_ENABLED`                                                   | `true`  | See [below](#scheduler_enabled-endpoint-details)                                 | read/write  |
 | [`wait_running_jobs`](../Deploy-Settings.md)                          | `true`  | `GET /tasks` (Swarm), `GET /containers/json` (Compose)                                  | read        |
-| [REST API](../Endpoints/REST-API.md) (`API_SECRET` set)               | off     | `GET /info`, `GET /containers/json`, `POST /containers/{id}/start|stop|restart`, Swarm: `GET /services`, `POST /services/{id}/update` | read/write  |
+| [REST API](../Endpoints/REST-API.md) (`API_SECRET` set)               | off     | See [below](#rest-api-endpoint-details)                                         | destructive |
+
+#### `SCHEDULER_ENABLED` endpoint details
+
+- Default (`execution_mode: restart`): `POST /containers/{id}/restart`; Swarm: `GET /services/{id}`, `POST /services/{id}/update`
+- `execution_mode: one_off`: `POST /containers/create`, `POST /containers/{id}/wait`, `POST /containers/{id}/start`; Swarm: `POST /services/create`, `DELETE /services/{id}`, `GET /tasks`
+- `stop_services`: `POST /containers/{id}/stop|start`; Swarm: `POST /services/{id}/update`, `GET /tasks`
+
+#### REST API endpoint details
+
+- Compose project management: `GET /containers/json`, `POST /containers/{id}/start|stop|restart`, `DELETE /containers/{id}`, `DELETE /networks/{id}`, `DELETE /volumes/{name}`, `DELETE /images/{name}`
+- Swarm stack/service management: `GET /services`, `POST /services/{id}/update`, `DELETE /services/{id}`, `DELETE /configs/{id}`, `DELETE /secrets/{id}`
 
 !!! danger "Image pruning in Swarm mode bypasses the proxy"
     In Swarm mode, `prune_images` runs `docker image prune` inside a global job service that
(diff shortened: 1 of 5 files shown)


wiki/docs/Advanced/Docker-API-Permissions.md
```diff
@@ -0,0 +1,238 @@
...
+
+Doco-CD does not call these endpoints, so you can safely deny them:
+
+`POST /auth`, `POST /build`, `POST /commit`, `POST /containers/{id}/exec`, `POST /exec/{id}/start`,
+`GET /containers/{id}/logs`, `GET /containers/{id}/top`, `POST /containers/{id}/pause`,
+`GET /plugins`, `POST /session`, `GET /system/df`, `POST /swarm/init`, `POST /swarm/join`
+
+!!! note "Building images"
+    Doco-CD does not build images by default. If any of your services use `build:` in their
+    compose file, you additionally need `POST /build` (and `POST /session` when BuildKit is used).
```
kimdre: This should be its own section or moved to the **Optional features** section. Users may overlook this since it's in the **Never used** section.

kimdre (changes requested):
Thank you for your PR! Please check my feedback to the changed docs. :)

kimdre:
I moved the required permissions for building images to a new section and added some missing permissions that are needed for the job scheduler and REST API.

kimdre (approved):

PR #1629, opened 2026-08-05
fix(swarm): stop recreating all tasks of a stack on every deployment

### LINKED ISSUE #1153: Only deploy service when changed in swarm mode

nosovk:
Draft, because this changes where deployment metadata lives in Swarm mode and has a few consequences that are worth agreeing on before merging. Details and open questions at the bottom.

Refs #1153. Measurements and the full analysis are in [this comment](https://github.com/kimdre/doco-cd/issues/1153#issuecomment-5195958432).

## Problem

Two separate places wrote deployment metadata into the task template:

1. `addSwarmServiceLabels` wrote 16 labels to `services.X.labels`, which `convert.Service` maps to `TaskTemplate.ContainerSpec.Labels`.
2. `addSwarmVolumeLabels` wrote 12 labels to `volumes.X.labels`, which `convert.Volume` maps to `TaskTemplate.ContainerSpec.Mounts[].VolumeOptions.Labels`.

Both are part of the task template. `cd.doco.deployment.timestamp` is recomputed on every deployment regardless of whether anything changed, so swarmkit's `IsTaskDirty` saw a different `TaskSpec` every time and replaced the tasks of every service in the stack.

Verified in isolation on a single-node swarm with a plain nginx service:

| Change | Result |
|---|---|
| `--container-label-add` | task `3aknjsxfvomp` → `9cvv5kj86661`, **recreated** |
| `--label-add` | task `9cvv5kj86661`, **unchanged** |

## Fix

Metadata now goes to `s.Deploy.Labels` → `ServiceSpec.Labels`, which is not part of the task template. The separate `serviceLevelLabels` set is gone — its five labels were already a subset of the metadata now written to the service.

For volumes, the four unstable labels (`Timestamp`, `CommitSHA`, `Trigger`, `Version`) are dropped. This is safe: volumes are looked up by `com.docker.stack.namespace`, and `userVolumeLabels` already discards everything prefixed `cd.doco.` before comparing volume configs, so nothing read them. Three parameters of `addSwarmVolumeLabels` became unused and were removed.

Config and secret labels are converted into separate `ConfigSpec`/`SecretSpec` objects rather than into the task template, so they are left alone.

**Backward compatibility.** Stacks deployed by earlier versions carry their metadata in the container spec. `SwarmServiceLabels()` merges both sets with the service spec taking precedence, and is used at all four read sites. Without it, `stage_3_destroy.go` would refuse to destroy doco-cd's own stacks with `ErrNotManagedByDocoCD`, and `getDeployStatus` would lose track of what is deployed. Verified against a service created the way older versions did — metadata is still read correctly.

**Job labels are exempt from the merge.** `cd.doco.job.*` labels are authored by the user as `services.X.labels`, so they belong to the task template, which is also where `isScheduledServiceSpec` and `scheduledRestartReplicas` read them. Merging them from the service spec would let the scheduler discover services that the deploy path never set up as jobs — in restart mode that would scale a `replicas: 3` service down to its restart replica count. `SwarmJobLabels()` keeps deployment metadata from the service spec but takes job configuration from the task template only.

## Verification

The idempotency test asserts both halves of the invariant: that the tasks survive a redeployment, **and** that the metadata was actually updated. Checking only the former would pass if `addSwarmServiceLabels` stopped writing anything at all — confirmed by temporarily removing the write, which the test now catches.

Each test was checked against a deliberately reintroduced bug to confirm it fails for the right reason. `TestSwarmJobLabels` caught a real bug in the first version of `SwarmJobLabels`, where a job label present in both places took the service spec value instead of the task template one.

Dumping the full service specs before and after a redeployment now shows `task_template_diffs=0` for every service, with only `/Labels/cd.doco.deployment.timestamp` changing.

`make fmt` reports 0 issues. `internal/docker`, `internal/reconciliation`, `internal/scheduler` and `internal/stages` all pass against a live swarm.

## Consequences worth agreeing on before merge

**1. Auto-discovery cleanup in Swarm starts working, and it deletes things.**

`cleanupObsoleteAutoDiscoveredContainers` filters services by `cd.doco.deployment.auto_discovery` through `ServiceList`, which only matches service spec labels. That label was never a service label, so the filter always returned nothing and the cleanup was dead code in Swarm mode. It now works.

That is a fix, but it activates a destructive path silently. `Delete` defaults to `true` and `RemoveImages` defaults to `true`, so obsolete auto-discovered stacks that have been sitting in a cluster untouched will start being removed. This deserves a release note at minimum. If you would rather it not switch on implicitly, I am happy to gate it or log a warning listing the candidates on the first run instead.

**2. One-off task recreation on upgrade.** Removing the labels from the task template is itself a change to the task template, so the first deployment after upgrading recreates the tasks once. Unavoidable; from the second deployment on it is quiet. Also worth a release note.

**3. `cd.doco.*` labels disappear from the containers on worker nodes.** They are now only visible via `docker service inspect` on a manager. Nothing inside doco-cd depends on them, but external tooling might — `docker ps --filter label=cd.doco.deployment.name=...`, log shippers, dashboards. I did not find this documented as a contract anywhere, so I treated it as acceptable, but flagging it in case you consider it one.

**4. The REST API response shape changes.** `GET /stacks` and `GET /stacks/{name}` return raw service JSON, so metadata moves from `Spec.TaskTemplate.ContainerSpec.Labels` to `Spec.Labels` for consumers.

## Found nearby, not fixed here

`getRunningScheduledJobServices` (`compose.go:917`) reads `api.ServiceLabel` from `ContainerSpec.Labels`, but `com.docker.compose.service` is never set on the swarm path — neither `addSwarmServiceLabels` nor `convert.Service` writes it. So `wait_running_jobs` appears to be a no-op in Swarm mode. This predates the change and I left it alone, but it is in the same area and I can open a separate issue if useful.

## Note

@zechmeister offered to take this on in the issue, and I do not want to step on that — happy to close this if they would rather do it. My commit is also unsigned, contrary to `CONTRIBUTING.md`, as I do not have a signing key set up on this machine; I can re-sign before merge.


zechmeister:
@nosovk don't worry about my initial draft. Very happy for you taking over!

kimdre:
Thank you for your PR! I haven't reviewed everything yet, but would it be possible to have a least some basic `cd.doco.X` labels on containers in Swarm mode?

nosovk:
Done in 5308cf5d. Containers now keep the stable subset of the metadata as container labels: `metadata.manager`, `deployment.name`, `deployment.working_dir`, `deployment.config.target`, `deployment.target.ref`, `source`, `source.name`, `source.url` — so `docker ps --filter label=cd.doco.deployment.name=...` keeps working on worker nodes.

The labels that change between deployments (`timestamp`, `target.sha`, `trigger`, `compose.sha`, `config.sha`, `metadata.version`, auto-discovery flags) stay service-spec only, since writing them to the task template is exactly what caused the task recreation. `TestAddSwarmServiceLabels_UsesServiceLevelLabels` now asserts both halves of the split, and `TestDeploySwarmStackIsIdempotent` still passes against a live swarm, confirming the stable subset does not dirty the task template.

One note: `working_dir` and `target.ref` are stable across ordinary redeployments but do change if the deployment is renamed or re-pointed at another ref — that single deployment will recreate tasks once, which seems acceptable since the config actually changed. If you would rather keep the container set minimal (`manager` + `deployment.name` + source labels only), happy to trim it.

nosovk:
Follow-up in 5c8c703a after a self-review pass, one correction to my previous comment:

**`source.url` is removed from the container-label subset.** Its value is not stable for the same stack: a webhook deploy stamps `html_url` (`https://github.com/owner/repo`) while a poll deploy stamps the raw clone URL (`https://github.com/owner/repo.git`), see `internal/stages/stage_1_init.go:243` vs `internal/webhook/payload.go:29`. In a webhook+poll-fallback setup the label would flip on every trigger-type switch and recreate all tasks — the exact bug this PR fixes. It stays available as a service label; `source.name` (`owner/repo`, derived identically in both paths) remains on containers for identification.

Also in this commit:
- The stable metadata map is extracted into `stableSwarmMetadataLabels`, shared by the service and volume label writers, so the stability invariant is defined in one place.
- `SwarmJobLabels` now exempts `cd.doco.job.last_run`/`next_run` from the task-template-only rule: those are runtime bookkeeping written by doco-cd itself to the service spec (`RerunJobService`), not user job config, so the scheduler can surface `last_run_at` for swarm jobs.
- The unit test now has an exhaustiveness check: any `cd.doco.*` container label not in the reviewed stable list fails the test, so an unstable label cannot slip back into the task template unnoticed.

`TestDeploySwarmStackIsIdempotent` and `TestDeploySwarmStack` re-verified against a live swarm; lint clean.

kimdre:
I had some time to review and test the changes.

Regarding the other change consequences:
1.  **Auto-discovery cleanup in Swarm starts working, and it deletes things**: It probably makes sense to set the default for the autodiscovery `Delete` option to `false` and let users enable it explicit to avoid unwanted stack deletions after updating doco-cd. Either we change it in this PR or a follow-up PR for better documentation. Feel free to give me feedback what you think. :)
2. **One-off task recreation on upgrade**: The unavoidable redeploy is fine for me. I will document this behavior in the release notes.
3. **`cd.doco.*` labels disappear from the containers on worker nodes**: Fixed in your commit above
4. **The REST API response shape changes**: The changed REST API response is fine for me. I will also document it in the release notes.

kimdre (approved):

PR #1695, opened 2026-08-17
fix(compose): fail start wait on crashlooping services

dennislapchenko:
Found this live on v0.108.0. Service that crashes right after start (misconfigured env, bot without required secret) was marked as successful deployment, then next poll saw the dead container as drift and redeployed - endless loop, success notification to telegram every ~30s.

Root cause in `waitForStartedServices`: it returns success on first sample where all services are `running`. Crashlooping container spends most of each cycle in a short running window, so one lucky 1s-tick sample wins. And `restarting` state is not handled at all in `assessStartedServiceStates` - service just stays in waiting list instead of failing.

Fix, two parts:
- `restarting` is terminal now: it only exists after container already died and restart policy kicked in, so right after deploy it always means crash-on-start
- readiness must hold 3 consecutive samples before success, so the short running window doesnt win the race. Costs 2 extra seconds per deploy, nothing else changes

Added a lister seam (`waitForStartedServicesWith`) so wait loop is unit-testable without docker daemon. Tests cover crashloop failure, streak reset on a dip, and stable success taking exactly 3 samples.

### DIFF
--- a/internal/docker/compose.go
+++ b/internal/docker/compose.go
@@ -1858,7 +1858,9 @@ func assessStartedServiceStates(containers []api.ContainerSummary, targetService
 			case "unhealthy":
 				status.unhealthy = true
 			}
-		case "exited", "dead":
+		// "restarting" exists only after a container died and restart policy
+		// kicked in - right after deploy that is a crash, not a slow start
+		case "restarting", "exited", "dead":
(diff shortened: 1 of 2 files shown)


kimdre:
Thanks :)

PR #1700, opened 2026-08-18
fix(notification): stop repeating and duplicating failure notifications

dennislapchenko:
## Problem

A failure notification repeats as often as its trigger. A poll job runs on its interval, a broken deployment fails the same way every run, so one fault gives one message per poll until somebody fixes it.

On my setup the registry token expired while the daemon was running. Result: **~240 identical messages in two hours**, and each poll sent **two** of them, because the stack reports its own failure and then the poll handler reports the same error again under its own title:

```
❌ Deployment Failed
app-dev -> app @ refs/heads/main (463450d) in 1.704s — failed to pull images: ... token has expired

❌ Poll Job failed
app-dev -> @ refs/heads/main in 0s — deployment failed: failed to pull images: ... token has expired
```

Nothing in the message changes between repeats, so every one after the first carries no information.

## Changes

**De-duplicate failures per stack.** Key is repository + target + stack + context, fingerprint is title + message. The first failure is sent, unchanged repeats are skipped, and one reminder goes out per `APPRISE_NOTIFY_REPEAT_INTERVAL` (new, default `1h`) so a long outage does not disappear. A different error on the same stack is sent immediately - that is new information. A **success** notification for that stack clears the state, so the existing `Deployment completed` is the recovery signal and no new notification type is needed. Set the interval to `0` to get the old behaviour back.

The fingerprint deliberately ignores metadata: job id and duration differ on every attempt, so including them would make every repeat look new.

**Report one fault once.** `StageManager.NotifyFailure` now returns the error marked as already reported, and the poll and webhook handlers skip their own notification for such an error. Logs, metrics and HTTP responses are unchanged - only the second message is dropped.

Suppression state is in memory and per stack, so a restarted daemon sends the next failure again.

## Tests

`internal/notification/failure_repeat_test.go` covers suppression of an unchanged repeat, the reminder after the interval, a changed fault and another stack passing through, clearing on success, the `0` opt-out, and the error marker. Existing `TestNotifyFailure*` extended for the returned marked error.

`go build`, `go vet`, `golangci-lint run` and the touched package tests are green.

## Docs

`wiki/docs/Advanced/Notifications.md`: the new setting in the table plus a "Repeated failures" section describing the behaviour.

### DIFF
--- a/internal/notification/failure_repeat.go
+++ b/internal/notification/failure_repeat.go
@@ -41,6 +41,10 @@ func SetFailureRepeatInterval(interval time.Duration) {
 	defer failureMu.Unlock()
 
 	failureRepeatInterval = interval
+
+	if interval <= 0 {
+		lastFailures = map[string]failureRecord{}
+	}
 }
 
 // failureKey identifies the thing that failed. The stack is what an operator
@@ -70,6 +74,8 @@ func shouldSendFailure(key, fingerprint string, now time.Time) bool {
 		return true
 	}
 
+	pruneFailures(now)
+
 	last, found := lastFailures[key]
 	if found && last.fingerprint == fingerprint && now.Sub(last.sentAt) < failureRepeatInterval {
 		return false
@@ -78,6 +84,18 @@ func shouldSendFailure(key, fingerprint string, now time.Time) bool {
 	lastFailures[key] = failureRecord{fingerprint: fingerprint, sentAt: now}
 
 	return true
+}
+
+// pruneFailures drops records that cannot suppress anything anymore. Past the
+// repeat interval the next failure is sent whatever the record says, so keeping
+// it only grows the map for stacks that are renamed, removed or fixed without a
+// success notification. Caller holds failureMu.
+func pruneFailures(now time.Time) {
+	for key, record := range lastFailures {
+		if now.Sub(record.sentAt) >= failureRepeatInterval {
+			delete(lastFailures, key)
+		}
+	}
 }
 
 // clearFailure forgets the last failure of a stack, so the next one is sent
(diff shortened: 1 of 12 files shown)


internal/notification/failure_repeat.go
```diff
@@ -0,0 +1,90 @@
...
+	if failureRepeatInterval <= 0 {
+		return true
+	}
+
+	last, found := lastFailures[key]
+	if found && last.fingerprint == fingerprint && now.Sub(last.sentAt) < failureRepeatInterval {
+		return false
+	}
+
+	lastFailures[key] = failureRecord{fingerprint: fingerprint, sentAt: now}
```
kimdre, line 84: There might be a potential memory leak in `lastFailures` because some old entries never get removed.

internal/notification/notified_error.go
```diff
@@ -0,0 +1,34 @@
...
+
+	return &notifiedError{err: err}
+}
+
+// WasNotified reports whether a failure notification was already sent for err.
+func WasNotified(err error) bool {
+	var notified *notifiedError
+
+	return errors.As(err, &notified)
+}
```
kimdre, line 38: Clarify in code comment that `WasNotified()` only works if errors are wrapped with `%w`, not reconstructed.

kimdre (changes requested):
Thanks for the PR! I found some minor things that should be improved.

internal/notification/failure_repeat.go
dennislapchenko: Done — records past the repeat interval are pruned now.

internal/notification/notified_error.go
dennislapchenko: Done.

kimdre (approved):

PR #1706, opened 2026-08-18
docs(examples): add full working examples for common setups

### LINKED ISSUE #1453: Full examples section
### Description

Seems like a great project, but struggle to find fully working examples, if you could add those to the repo/docs would be of great help in adoption. 
Something like "best practices" or full working examples for various scenarios.
[doco-cd_tests](https://github.com/kimdre/doco-cd_tests) is kind of that? but not sure its well documented

### Use Case

Want to start using, but features and schemas are a bit obscure.

### Additional Information

_No response_

dennislapchenko:
Follow-up to #1453 — the examples I promised there, distilled from three real doco-cd installs I run.

Adds `examples/` with three complete setups. Each replicates every repo it needs: `app-repo/` (or `deployments-repo/`) is what goes to git, `server/` is what goes on the VM. Small README per example explains the pattern and how to try it.

- **single-repo-single-env** — app repo carries its own compose + `.doco-cd.yml`, one VM polls it. Shows image tag pinned in git (bump = release), secrets kept on VM via `PASS_ENV`, poll-only daemon with no open ports, mounted-config auto-apply.
- **single-repo-two-envs** — same repo, staging + prod via poll `target:` isolation. Shows per-stack image pins (stateful side not recreated on web-only release), why reconciliation should not watch `die` when a one-shot migrate container exists, external network between stacks, apprise notifications with a compact body template.
- **deployments-repo** — one central repo deploys many apps to many VMs. Per-VM target files pin every version (nothing floats), per-env compose stubs `include:` the app repo's compose pinned at a sha, secrets SOPS-encrypted in the repo and decrypted by the daemon with an age/KMS key.

All configs are things running in production for me right now, only names sanitized. YAML validated, composes render with `docker compose config`.

Docs change is minimal per the idea to keep examples in the repo: one tip box in Getting Started linking to the directory. 

I am sure you have usages which are different from this. And maybe I even have some wrong patterns somewhere? IF so this PR is a good opportunity to fix that and publish these first examples. Then you could add yours following the idea

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,8 @@ ed25519
 
 *.env
 .env
+# examples commit their env files on purpose (cleartext non-secret config)
+!examples/**/*.env
 
 cover.out
 cover.html
--- a/examples/README.md
+++ b/examples/README.md
@@ -3,16 +3,16 @@
 Full working setups for common doco-cd scenarios. Each example replicates every repo it needs:
 
 - `app-repo/` or `deployments-repo/` — what you commit to Git.
-- `server/` — what you put on the VM next to Docker (doco-cd itself).
+- `server/` — what you put on the Docker host (doco-cd itself).
 
 | Example | Pattern | Pick it when |
 |---|---|---|
-| [single-repo-single-env](single-repo-single-env/) | App repo carries its own compose + `.doco-cd.yml`. One VM polls it. | You have one app and one server. Start here. |
-| [single-repo-two-envs](single-repo-two-envs/) | Same repo, two deploy configs. Each VM selects its own via poll `target:`. Adds notifications. | You have staging and prod of the same app. |
-| [deployments-repo](deployments-repo/) | One central repo deploys many apps to many VMs. Version pins per VM, compose files pulled from app repos, secrets encrypted with SOPS. | You have a fleet of apps/VMs and want one place that states what runs where. |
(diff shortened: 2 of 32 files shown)


kimdre (commented):
Sorry for taking so long. In your documentation you always mention VMs but I don't really like this wording. Many users don't use a VM for their Docker host and sometimes you also use *VM* to describe a doco-cd instance which can be confusing.

I prefer the word *host* or something similiar for cases where the text is about a docker host and *doco-cd instance* for cases where it's about what doco-cd does (on a docker host).

dennislapchenko:
No problem, thanks for looking.

Agree, "VM" was wrong word. Dropped it everywhere — in `examples/` and in the Getting Started tip. Now follows what wiki already uses: *host* / *Docker host* for the machine, *doco-cd instance* for what doco-cd does on it. Placeholder `.doco-cd.<vm>.yml` became `.doco-cd.<host>.yml` too. Also killed two "on a box" leftovers, same problem.

Directory is still named `server/`, I did not rename it — rename touches every path in every README. If you want `host/` instead, say so and I do it.

Re-checked: yaml parses, composes still render with `docker compose config`, `go build ./...` and `go vet ./...` clean.

kimdre (approved):
Thanks, now it's good :)

dennislapchenko:
> Thanks, now it's good :)

What do you think overall, am I using doco-cd correctly? can something be improved perhaps?

PR #1711, opened 2026-08-19
fix(deploy): retry deployments that failed after containers were created

dennislapchenko:
Implements option 1 from discussion #1702: endless retries until success, outcome recorded separately from the commit labels.

## What changed

- On a failure in the deploy stage or later, doco-cd records the failed attempt (commit, stage, error, timestamp) in memory, replaced on each attempt. Daemon restart drops the record and the regular label comparison applies again.
- Pre-deploy checks the marker before the skip logic. Record present = the labels lie, deployment proceeds. This repeats on every poll/webhook until one attempt succeeds, then the record is removed. Successful destroy removes it too.
- Retry of a failed **deploy stage** force-recreates the whole project. Without that the retry is a false success: unchanged compose config is a no-op for `up`, containers stay running, the failed `post_start` hook never runs again, and the record would be cleared with migrations still not applied. Recreation makes the retry actually re-run the failed part. Failures of post-deploy/cleanup (deployment itself went through) retry without recreation.
- Failures of init/pre-deploy are not recorded - they happen before anything is mutated and retry naturally, like the pull-failure path always did.
- Every failed attempt still reports a failure notification, so with the de-duplication from #1700 the cost is one message plus periodic reminders, exactly as discussed.

## Why not the label

Labels are not implementable for this without recreating containers: docker cannot change labels on an existing container (`docker update` only takes resources and restart policy), and compose writes labels only when it creates containers. A post-stages label write would need exactly the second recreate pass that made us both reject option 2. So the record is in-memory now, per review.

## e2e proof

Second commit adds a black-box e2e harness (`test/e2e/`) plus a scenario for exactly this bug. The harness builds doco-cd from the branch, runs it against a local anonymous git server, scenarios push commits and assert on containers and daemon state. One entry point for humans and CI: `./test/e2e/run.sh`, also wired to run on PRs via a new workflow.

The `failed-deploy-retry` scenario: v1 with a `post_start` hook exiting 1 -> asserts the deploy fails, the unchanged commit is retried with a full recreate (container id changes), then a fix commit -> deploy succeeds, record clears (proven by the first "no changes" skip in the log) and the stack stays untouched for two more poll cycles.

Verified both ways: passes on this branch, fails on current main exactly at the "retry detected on unchanged commit" step while the daemon logs `no changes detected, skipping deployment` every poll - the very lines from the discussion report.

Scenarios are bash on purpose - e2e here is "push, wait, assert" and bash reads as the spec. Go test funcs would work too, just heavier per scenario - lmk if you want that instead. And if e2e harness is worth keeping, lmk if it should come as its own PR.


dennislapchenko:
sidenote, docker compose guys merged this: https://github.com/docker/compose/pull/14088
So soon we will have the ability to see actual errors on hooks

kimdre:
Thank you for your work. :)

Since the docker compose team is already [working](https://github.com/docker/compose/pull/14091) on improvements for hooks and also merged your PR there, maybe it's best to wait for that release iteration to avoid adding more complexity to this project for something that might be fixed or at least simplified upstream soon. What do you think?

Regarding the marker file, would it be sufficient to keep this state in-memory so we don't have to write a file?

dennislapchenko:
> Thank you for your work. :)
> 
> Since the docker compose team is already [working](https://github.com/docker/compose/pull/14091) on improvements for hooks and also merged your PR there, maybe it's best to wait for that release iteration to avoid adding more complexity to this project for something that might be fixed or at least simplified upstream soon. What do you think?
> 
> Regarding the marker file, would it be sufficient to keep this state in-memory so we don't have to write a file?

Waiting on compose guys - I am not sure it will help doco-cd with failed deployment recreations? their improvements are just about what is becoming better visible

In-memory instead of marker file - probably making it lighter is a good choice, understanding that doco-cd rollout resets this, but is probably cheap.

dennislapchenko:
What do you think of the e2e harness?
I think you kind of have same in your git submodule, but you mentioned it's for you.
Some kinda real e2e testing "framework" should be part of PR CI I believe.

kimdre:
> > Thank you for your work. :)
> > Since the docker compose team is already [working](https://github.com/docker/compose/pull/14091) on improvements for hooks and also merged your PR there, maybe it's best to wait for that release iteration to avoid adding more complexity to this project for something that might be fixed or at least simplified upstream soon. What do you think?
> > Regarding the marker file, would it be sufficient to keep this state in-memory so we don't have to write a file?
> 
> Waiting on compose guys - I am not sure it will help doco-cd with failed deployment recreations? their improvements are just about what is becoming better visible

Yes but this could potentially simplify the implementation and make it less hacky since currently it's hard to track failed compose hooks.

test/e2e/harness/gitserver/Dockerfile
kimdre, line 1: Could we use the official git or nginx image as the base image to speed up the build?

kimdre (changes requested):
I really like the end-to-end implementation. 

Could you add a test-e2e command to the makefile?

dennislapchenko:
Done: failure record now in-memory instead of file, `make test-e2e` added, gitserver on official nginx alpine base.

On waiting for compose: their improvements make hook failures visible and better structured, but the retry/recreate decision stays on doco-cd side either way - so waiting alone doesn't close #1702. Counter-offer: lets add experiments concept, terragrunt style. `EXPERIMENTS` env (comma list), each experiment documented with what it does + feedback link + stabilization criteria, on GA the flag becomes harmless no-op. This retry becomes the first experiment, default off - zero behavior change for everyone else, PR unblocks, and when the compose release lands we revisit: simplify on top of it and promote, or drop. Happy to do that as follow-up commit here or separate PR.

e2e stays in this PR imo - split out it would be a framework with empty scenario set. wdyt?

.github/workflows/e2e.yaml
kimdre, line 1: I would prefer to have the e2e job in the existing test.yaml file so all tests are grouped in the "Tests" workflow if possible.

kimdre (changes requested):

kimdre:
On a different thought, you are right. I will approve your implementation, no need to wait.

By the way, is there a reason why you chose to write the e2e tests using shell scripts rather than something like [Testcontainers for Go](https://github.com/testcontainers/testcontainers-go) with a special test flag like `go test -tags=e2e ./...`?

.github/workflows/e2e.yaml
dennislapchenko: Done, moved to test.yaml as e2e job. Worth knowing: it now runs on push to main too, since test.yaml triggers there.

dennislapchenko:
Choice was mostly about what is under test. doco-cd runs in container and creates its own compose stacks through mounted docker socket — those stacks live outside testcontainers lifecycle, Ryuk will not reap them, and assertions are `docker ps` / `docker logs` anyway. Harness is plain compose file, same thing a user writes, so it doubles as documentation. Plus no new dep, and no chance e2e leaks into `go test ./...`.

Fair points against shell though: bash portability (already had to dance around `sed -i` on mac), weak assertions, no reuse of existing Go helpers.

Middle ground: keep `harness/compose.yaml` as is, drive it from Go with `//go:build e2e` + testcontainers compose module, scenarios become Go tests, entrypoint `go test -tags=e2e ./test/e2e/...`. Want to try? as follow-up or here?

kimdre:
> Middle ground: keep `harness/compose.yaml` as is, drive it from Go with `//go:build e2e` + testcontainers compose module, scenarios become Go tests, entrypoint `go test -tags=e2e ./test/e2e/...`. Want to try? as follow-up or here?

Sounds good to me. We could either use testcontainers or extend the `test` helper package at `internal/test`.

kimdre (approved):

kimdre:
I will merge this PR for now. The improvement to the e2e tests can be done in a follow-up PR. Thanks again for your work! :)

PR #1733, opened 2026-08-24
feat(notification): name services whose image moved on tag bump

dennislapchenko:
`.ChangedServices` stays empty on plain tag bump. Two sources fill it — file-mount force-recreates, and digest drift behind `force_image_pull` (#1692). Tag bump in deploy config hits neither, so on normal deploy nothing is named.

Adds `DeployedServicesWithChangedImageRefs`: configured image ref against deployed one. Local only, no registry round trip, so it can run on every deploy where digest path did not answer already. Refs are normalized first, `nginx` and `docker.io/library/nginx:latest` do not count as change.

Runs after the skip decision, and error is logged, not returned — this fills notification only, it must not decide if deploy happens.

Cases:

- first deploy of stack — nothing, stack message already says it came up
- deployed ref cannot be determined — service skipped
- service new to already deployed stack — reported

15 unit tests. `go build`, `go vet` and `golangci-lint run ./...` clean.

### DIFF
--- a/internal/docker/images.go
+++ b/internal/docker/images.go
@@ -619,7 +619,7 @@ func DeployedServicesWithChangedImageDigests(ctx context.Context, dockerCli comm
 // compare equal. A digest-pinned reference is already exact and is left alone; only a
 // tag-less reference needs `:latest` filled in.
 //
-// An unparseable reference is returned untouched: comparing two raw strings is still
+// An unparsable reference is returned untouched: comparing two raw strings is still
 // better than dropping the comparison entirely.
 func normalizeImageRef(ref string) string {
 	if ref == "" {
--- a/internal/docker/images.go
+++ b/internal/docker/images.go
@@ -525,6 +525,7 @@ func PullImages(ctx context.Context, dockerCli command.Cli, projectName string) 
 var (
 	registryDigestLookup             = registryDigestForRef           // registryDigestLookup fetches registry digests for image refs, can be overridden in tests
 	deployedServiceDigestLookup      = getDeployedServiceImageDigests // deployedServiceDigestLookup fetches deployed service image digests, can be overridden in tests
+	deployedServiceRefLookup         = getDeployedServiceImageRefs    // deployedServiceRefLookup fetches deployed service image references, can be overridden in tests
 	registryDigestHeadLookup         = registryDigestForRefViaHEAD
 	registryDigestDistributionLookup = registryDigestForRefViaDistributionInspect
 )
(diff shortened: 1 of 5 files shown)


internal/docker/images.go
```diff
@@ -610,3 +611,156 @@ func DeployedServicesWithChangedImageDigests(ctx context.Context, dockerCli comm
...
 }
+
+// normalizeImageRef returns a comparable form of an image reference.
+//
+// Docker reports a deployed image in whatever form it was pulled while compose reports
+// what the project file says, so `nginx` and `docker.io/library/nginx:latest` have to
+// compare equal. A digest-pinned reference is already exact and is left alone; only a
+// tag-less reference needs `:latest` filled in.
+//
+// An unparseable reference is returned untouched: comparing two raw strings is still
```
kimdre: Please fix the spelling error: `unparseable ==> unparsable`

internal/docker/images_test.go
```diff
@@ -579,3 +579,152 @@ func requireRegistryIntegrationTestGate(t *testing.T) {
...
+		ref  string
+		want string
+	}{
+		{name: "empty stays empty", ref: "", want: ""},
+		{name: "bare name gets registry and tag", ref: "nginx", want: "docker.io/library/nginx:latest"},
+		{name: "tagged name gets registry", ref: "nginx:1.27", want: "docker.io/library/nginx:1.27"},
+		{name: "fully qualified is stable", ref: "docker.io/library/nginx:latest", want: "docker.io/library/nginx:latest"},
+		{name: "private registry is preserved", ref: "637423286173.dkr.ecr.eu-central-1.amazonaws.com/login-be:20260824-e5136680", want: "637423286173.dkr.ecr.eu-central-1.amazonaws.com/login-be:20260824-e5136680"},
+		{name: "digest pin is left exact", ref: "nginx@sha256:0000000000000000000000000000000000000000000000000000000000000000", want: "docker.io/library/nginx@sha256:0000000000000000000000000000000000000000000000000000000000000000"},
+		{name: "unparseable ref is returned untouched", ref: "NOT A REF", want: "NOT A REF"},
```
kimdre: Please fix the spelling error: `unparseable ==> unparsable`

kimdre (changes requested):
Thanks for the PR. Please fix the spelling errors before I approve the changes. :)

dennislapchenko:
Done!

internal/docker/images.go
```diff
@@ -610,3 +611,156 @@ func DeployedServicesWithChangedImageDigests(ctx context.Context, dockerCli comm
...
+		}
+
+		if deployedRef != configuredRef {
+			logger.Info("service image reference changed",
+				slog.String("service", svc.Name),
+				slog.Group("image",
+					slog.String("deployed", deployedRef),
+					slog.String("configured", configuredRef),
+				),
+			)
```
kimdre: I may have found a bug here, Swarm services are always flagged as image-changed.
Line 750 compares the configured tag (for example, `nginx:1.27`) directly with Swarm’s deployed reference (`nginx:1.27@sha256:...`). `normalizeImageRef` intentionally retains canonical digest references, so these never match. As a result, any tagged Swarm service is incorrectly included in `ChangedServices` on unrelated redeployments. Compare name/tag while disregarding the deployed digest when the configured image is tag-based; retain digest comparison for explicitly digest-pinned configuration

kimdre (changes requested):

dennislapchenko:
Good catch! fixed

kimdre (approved):

kimdre:
Thanks!

PR #1748, opened 2026-08-26
feat(external-secrets): add support for interpolation in secret references

zyojv:
I’m using Doco-CD in two different environments: my lab and my production site. I deploy the same compose stacks in both environments, but the certificate common names differ between them.

Currently, this isn’t possible because the external secret values are fixed in `.doco-cd.yml`, and these files are packaged into OCI images.

For example, the lab environment currently uses:

```yaml
name: myapp
external_secrets:
  CERT: pki:pki:app.lab.acme.corp
```

While production uses:

```yaml
name: myapp
external_secrets:
  CERT: pki:pki:app.prod.acme.corp
```

This means I need to build or maintain different configurations for each environment, even though the Compose stack itself is identical.

## Proposed Solution

Allow environment variable interpolation in `external_secrets`, similar to Compose variable interpolation.

For example:

```yaml
name: myapp
external_secrets:
  CERT: pki:pki:app.${COMPOSE_STAGE:prod}.acme.corp
```

Then, `COMPOSE_STAGE` could be set differently for each Doco-CD deployment:

* Lab: `COMPOSE_STAGE=lab`
* Production: `COMPOSE_STAGE=prod`

This would allow the same `.doco-cd.yml` and OCI image to be used across both environments while keeping environment-specific values outside of the image.

The default value (`prod` in the example above) would also make the configuration behave sensibly when `COMPOSE_STAGE` is not explicitly set.

## Motivation

This would be particularly useful when the same OCI image and Compose stack are deployed across multiple environments where only environment-specific configuration, such as certificate names, domains, or other external secret references, differs.

**If there is already a supported way to achieve this without duplicating `.doco-cd.yml` files or OCI images, please point me in the right direction. I may be missing an existing feature or recommended approach.**


internal/secretprovider/types/secret_ref.go
```diff
@@ -125,3 +128,26 @@ func EncodeExternalSecretRefs(in map[string]ExternalSecretRef) (map[string]strin
...
+
+// InterpolateExternalSecretRefs expands Compose-style variables in external
+// secret references using the doco-cd process environment. It never invokes a
+// shell or uses deployment environment values.
+func InterpolateExternalSecretRefs(in map[string]ExternalSecretRef) (map[string]ExternalSecretRef, error) {
+	out := make(map[string]ExternalSecretRef, len(in))
+
+	for envName, ref := range in {
+		if ref.LegacyRef == "" {
+			continue
```
kimdre, line 162: Structured secret refs are dropped. This skips refs with empty `LegacyRef`, which includes every object-form `store_ref`/`remote_ref`. Webhook secrets therefore never reach the provider.

internal/secretprovider/types/secret_ref.go
```diff
@@ -125,3 +128,26 @@ func EncodeExternalSecretRefs(in map[string]ExternalSecretRef) (map[string]strin
...
+		}
+
+		value, err := template.SubstituteWithOptions(ref.LegacyRef, os.LookupEnv, template.WithoutLogging)
```
kimdre: `PASS_ENV=false` is bypassed. This line reads all process variables unconditionally. Repository config can inject sensitive doco-cd environment values into provider requests and ref-bearing errors.

internal/stages/stage_2_pre-deploy.go
```diff
@@ -124,7 +124,12 @@ func (s *StageManager) RunPreDeployStage(ctx context.Context, stageLog *slog.Log
 	if s.SecretProvider != nil && *s.SecretProvider != nil && len(s.DeployConfig.ExternalSecrets) > 0 {
 		stageLog.Debug("resolving external secrets", slog.Any("external_secrets", s.DeployConfig.ExternalSecrets))
 
-		encodedSecrets, err := secrettypes.EncodeExternalSecretRefs(s.DeployConfig.ExternalSecrets)
+		interpolatedRefs, err := secrettypes.InterpolateExternalSecretRefs(s.DeployConfig.ExternalSecrets)
```
kimdre: Interpolated PKI refs break revocation detection. This only uses expanded refs for resolution. Certificate labels retain `${...}`, causing OpenBao’s revocation parser to fail and the watcher to skip checks.

internal/stages/stage_2_pre-deploy.go
kimdre, line 1: Environment changes may not redeploy. OCI deployments skip on unchanged digest, while Git `pki-role` hashes normalize to the original template. Changing only an interpolation variable can leave the old secret or certificate deployed

kimdre (changes requested):
Thank you for the PR! The feature is solid in concept but needs some important fixes before merge:

- Preserve structured refs unchanged through interpolation (don't skip them).
- Write interpolated refs back into DeployConfig so all consumers see resolved values.
- Gate interpolation behind `PASS_ENV` or a dedicated flag, or explicitly document the security implication.
- Please add some tests with actual structured refs and pki-role templates.

In its current state this will break existing webhook deployments and cause OpenBao certificate rotation to fail silently.

zyojv:
@kimdre thanks for the feedback. I'll get to it as soon as my workday ends.

zyojv:
I've addressed all the issues, introduced `INTERPOLATE_EXTERNAL_SECRETS` instead for re-using `PASS_ENV`, and also tested it locally.

internal/stages/stage_2_pre-deploy.go
kimdre: The Git side is good now. Changing an env var updates the hash and triggers a redeploy.
OCI still has a problem though. It only checks if the image digest changed. So if you update an interpolation variable but the digest stays the same, it'll just skip the deploy and you'll be stuck with the old secrets or certs.
We would also need to add something to check for changes in environment variables, not just the digest.

kimdre:
There is still an issue with OCI, see my comment https://github.com/kimdre/doco-cd/pull/1748#discussion_r3881327108. There are also some linter errors that need to be solved.

internal/secretprovider/types/secret_ref.go
kimdre: One more thing here. If `${SECRET_ID}` is unset, interpolation returns an empty string. `EncodedReference` then mistakes the legacy scalar for a structured reference and sends:
`{"store_ref":"","remote_ref":null}`

The code should reject an empty interpolated result or preserve the reference variant explicitly.

kimdre (changes requested):
I noticed one more bug in `secret_ref.go`

kimdre (approved):

kimdre:
Thanks for your PR! :)

zyojv:
> Thanks for your PR! :)

Thank you for your great tool

PR #1765, opened 2026-08-28
fix(api): use snake case for compose responses

jkossis:
## Summary

- map Docker Compose project and container results into API-owned response types
- serialize every exposed field with explicit `snake_case` JSON names
- preserve the existing `content` envelope and response values

## Breaking change

This renames fields returned by `GET /v1/api/projects` and `GET /v1/api/project/{projectName}` from their dependency-defined PascalCase names to snake_case. Existing API consumers must update accordingly.

The Compose response structs come from `github.com/docker/compose/v5/pkg/api`, so Doco-CD cannot add tags to those dependency types directly; the response mapping keeps the wire contract under Doco-CD control.

## Test plan

- `GOTOOLCHAIN=local /tmp/go/bin/go test ./cmd/doco-cd -count=1`
- endpoint tests assert snake_case fields inside the `content` envelope and reject representative PascalCase fields

### DIFF
--- a/cmd/doco-cd/api_response.go
+++ b/cmd/doco-cd/api_response.go
@@ -1,0 +1,108 @@
+package main
+
+import (
+	"github.com/docker/compose/v5/pkg/api"
+	"github.com/moby/moby/api/types/container"
+)
+
+type projectResponse struct {
+	ID          string `json:"id"`
+	Name        string `json:"name"`
+	Status      string `json:"status"`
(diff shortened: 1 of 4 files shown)


kimdre:
The doco-cd API returns raw values from the upstream Docker API on purpose. Maintaining a copy of all returned fields just to change the casing is unnecessary work. A better solution would be to convert all JSON fields on the fly but this does not have priority for me at the moment.

PR #1790, opened 2026-09-02
Fix OCI deployment path reconstruction for scheduled runs and cert rotation

zyojv:
Scheduled runs and cert-rotation redeploys no longer have the original deployment context
around, so they rebuild the path to the deployment's source directory from its container
labels. 
That rebuild always assumed a Git source. For an OCI reference this kept the
`:latest` part in the folder name, but OCI artifacts are extracted into a folder without
the tag. So the rebuilt path pointed at a directory that doesn't exist. The deploy config
couldn't be reloaded and the scheduled run or rotation failed.


kimdre:
Thanks! :)

PR #1819, opened 2026-09-06
feat(openbao): expose full certificate chain as a _FULL companion variable

zyojv:
PKI references now always expose the full certificate chain as an companion environment (`CERT_FULL`) similar to `CERT_KEY`.

### DIFF
--- a/internal/docker/cert_rotation_labels.go
+++ b/internal/docker/cert_rotation_labels.go
@@ -26,8 +26,12 @@ const (
 )
 
 // pkiRoleKeySuffix is the suffix appended to a pki-role external secret's env
-// var name for its private key companion entry.
-const pkiRoleKeySuffix = secrettypes.PKIRoleKeySuffix
+// var name for its private key companion entry, and pkiFullChainSuffix the one for the full
+// certificate chain companion entry exposed for both pki and pki-role references.
+const (
+	pkiRoleKeySuffix   = secrettypes.PKIRoleKeySuffix
+	pkiFullChainSuffix = secrettypes.PKIFullChainSuffix
+)
 
 type deployedRotatableCertState struct {
 	Ref    string `json:"ref"`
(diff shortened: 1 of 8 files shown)


internal/docker/cert_rotation_labels.go
kimdre, line 325: Revoked certificates don't get detected for services using only `CERT_FULL`. 
Services consuming only `CERT_FULL` won't have revocation checks run on them. 
You need to add a check if the service uses the full chain value when deciding whether to track certificate state.

internal/secretprovider/openbao/pki.go
```diff
@@ -48,27 +112,78 @@ func GetCertSerial(ctx context.Context, client *api.Client, engineName, commonNa
...
+	cert, chain, err := readCert(ctx, client, engineName, serial)
+	if err != nil {
+		return "", "", err
+	}
+
+	if len(chain) == 0 {
+		chain, err = GetCAChain(ctx, client, engineName)
+		if err != nil {
+			return "", "", err
+		}
```
kimdre, line 133: Full certificate chains might be wrong in setups with multiple certificate authorities.

When reading a stored certificate that doesn't have its chain included, the code asks for "the default CA chain." But if your setup has multiple certificate authorities, that certificate might not belong to the default one. So you might get a chain from the wrong authority, making the `CERT_FULL` value invalid.

The system assumes all certificates come from the same authority, but this isn't always true.

Instead of asking for the default chain, figure out which authority actually issued the certificate and ask for *that* authority's chain.

kimdre (changes requested):
Thank you for the PR :) I found 2 issues in your code that need to get fixed.

zyojv:
Thanks for the feedback. I'll fix them as soon as I have time to spare. In the meantime I have converted the PR back to a draft

PR #1837, opened 2026-09-09
fix(encryption): keep decrypted files from every bind-mounted directory

### LINKED ISSUE #1836: SOPS-encrypted env_file values are passed to containers undecrypted when the project has a bind-mounted directory without encrypted files
### Description

When a compose project references a SOPS-encrypted dotenv file via `env_file` and also has more than one bind-mounted directory, the deployed containers sometimes receive the encrypted values (e.g. `POSTGRES_USER=ENC[AES256_GCM,data:...]`) even though the file in the working copy has been decrypted in place.

Root cause (internal/docker/compose_load.go, DecryptProjectFiles):
```go
    if info.IsDir() {
        decryptedFiles, err = encryption.DecryptFilesInDirectory(repoPath, v.Source)
        ...
        continue
    }
```

The result of each directory walk is assigned to `decryptedFiles` instead of appended. When a directory that contains no encrypted files (e.g. `./html`) is processed after the directory that contains the encrypted env file (e.g. `./.config`), the list becomes empty. The env file itself is already decrypted by then, so the final `DecryptFileInPlace` loop adds nothing either. `projectFilesDecrypted` ends up false, the project is not reloaded, and the environment resolved during the first `LoadProject` (read from the still-encrypted file) is used for the deployment.

Because `p.Services` is a map, the order of the directory walks is random, so the bug is intermittent. It does not occur when the project has only one bind-mounted directory. It occurs whenever the encrypted file on disk was (re)placed by a git sync, i.e. on the first deployment of a repository and on every deployment after a commit that changed the encrypted file. A retry on the next poll succeeds because the file is already decrypted on disk at that point (which also hides the bug when the failure is not detected by a healthcheck).

### Steps to reproduce the issue

```bash
1. Encrypt a dotenv file with SOPS (age) and commit it as .config/.env
2. Use the compose file below: one service uses env_file .config/.env, one service bind-mounts the .config directory, one service bind-mounts a directory that contains no encrypted files (./html)
3. Provide the age key to doco-cd via SOPS_AGE_KEY_FILE and let it deploy the repository (polling)
4. docker inspect <db container> --format '{{range .Config.Env}}{{println .}}{{end}}' | grep POSTGRES_USER
   -> POSTGRES_USER=ENC[AES256_GCM,data:...] (intermittent; repeat step 3 after a commit that changes .config/.env if it does not happen the first time)
5. The healthcheck that uses $POSTGRES_USER fails and the deployment ends with "dependency failed to start: container ...-db-1 is unhealthy"
6. On the next poll the retry succeeds, because .config/.env in the working copy is already decrypted
7. Remove the ./html bind mount (so that .config is the only bind-mounted directory): the problem no longer occurs
```

### doco-cd Version

v0.117.0 (ghcr.io/kimdre/doco-cd:0.117.0), Docker Engine 29.8.0, Compose plugin 5.5.1

### Docker Compose File

```yaml
services:
  db:
    image: groonga/pgroonga:latest-debian-18
    env_file:
      - .config/.env          # SOPS-encrypted (age), dotenv format
    healthcheck:
      test: "pg_isready -U $$POSTGRES_USER -d $$POSTGRES_DB"
      interval: 5s
      retries: 20

  backup:
    image: ghcr.io/team-shahu/misskey-backup:2026.08.16
    env_file:
      - .config/.env
    volumes:
      - .config:/app/config:ro   # bind-mounted directory that contains the encrypted file

  maintenance:
    image: caddy:2-alpine
    volumes:
      - ./html:/usr/share/caddy:ro   # bind-mounted directory without encrypted files

# .config/.env (before encryption)
# POSTGRES_USER=misskey
# POSTGRES_PASSWORD=...
# POSTGRES_DB=mk1
```

### Deployment Config

```yaml
name: example
reference: main
compose_files:
  - compose.yaml
timeout: 600
```

### Logs

```sh
# doco-cd
{"level":"info","msg":"deploying stack","deploy":{"recreate":{"forced_services":[],"mode":"force"},"reference":"main",...}}
{"level":"error","msg":"deployment failed","deploy":{"error":"failed to deploy stack example: failed to deploy stack: dependency failed to start: container example-db-1 is unhealthy",...}}
{"level":"warning","msg":"job completed with errors","elapsed_time":"2m37.227s",...}
# next poll
{"level":"warning","msg":"last deployment attempt failed, retrying",...}
{"level":"info","msg":"job completed successfully","elapsed_time":"28.446s",...}

# container environment while the deployment was failing
$ docker inspect example-db-1 --format '{{range .Config.Env}}{{println .}}{{end}}' | grep POSTGRES_USER
POSTGRES_USER=ENC[AES256_GCM,data:wT/l,iv:E3v+bzB5TRC2ytWReoKZJE31MDIvZI342UylVHinco0=,tag:...,type:str]

# the same file in the working copy at that time was already decrypted
$ grep -c 'ENC\[' /data/github.com/<org>/<repo>/.config/.env
0

# application inside a container that also read the value
level=fatal msg="Failed to create backup service: failed to create R2 storage: invalid R2 endpoint format: ENC[AES256_GCM,data:..."
```

chan-mai:
Fixes #1836

## Problem

`DecryptProjectFiles` assigns the result of each bind-mounted directory walk to `decryptedFiles` instead of appending to it:

```go
if info.IsDir() {
    decryptedFiles, err = encryption.DecryptFilesInDirectory(repoPath, v.Source)
    ...
    continue
}
```

When a directory without encrypted files (e.g. ./html) is walked after the directory that contains the encrypted env file (e.g. ./.config), the list becomes empty. The env file itself has already been decrypted by the earlier walk, so the final `DecryptFileInPlace` loop adds nothing either. `projectFilesDecrypted` ends up false, `LoadCompose` does not reload the project, and the environment resolved during the first `LoadProject` (read from the still-encrypted `env_file`) is deployed. Containers then receive values such as `POSTGRES_USER=ENC[AES256_GCM,...]`.

Because `p.Services` is a map, the walk order is random and the bug is intermittent. It shows up on the first deployment of a repository and after every commit that changes the encrypted file, i.e. whenever the file on disk was replaced by the git sync.

## Change

- Append the result of each directory walk to `decryptedFiles` instead of overwriting it.

## Test

- Added a case to `TestDecryptProjectFiles`: one service with a bind-mounted directory containing an encrypted env file followed by a bind-mounted directory without encrypted files (volumes are a slice, so the order is deterministic), and the same file referenced via `env_file`. The case fails on main (expected 1 decrypted file(s), got 0) and passes with this change.

### DIFF
--- a/internal/docker/compose_load.go
+++ b/internal/docker/compose_load.go
@@ -238,7 +238,7 @@ func DecryptProjectFiles(repoPath string, p *types.Project) ([]string, error) {
 				}
 
 				if info.IsDir() {
-					decryptedFiles, err = encryption.DecryptFilesInDirectory(repoPath, v.Source)
+					files, err := encryption.DecryptFilesInDirectory(repoPath, v.Source)
 					if err != nil {
 						if errors.Is(err, filesystem.ErrPathTraversal) {
 							continue
@@ -246,6 +246,8 @@ func DecryptProjectFiles(repoPath string, p *types.Project) ([]string, error) {
 
 						return decryptedFiles, fmt.Errorf("failed to decrypt files in bind mount directory '%s': %w", v.Source, err)
 					}
+
+					decryptedFiles = append(decryptedFiles, files...)
 
 					continue
 				}
--- a/internal/docker/compose_test.go
+++ b/internal/docker/compose_test.go
@@ -2705,6 +2705,47 @@ func TestDecryptProjectFiles(t *testing.T) {
 			},
 			expectedDecryptedFileBasenames: []string{},
 		},
+		{
+			name: "bind-mounted directory without encrypted files does not discard earlier results",
+			buildProject: func(t *testing.T, tmpDir string) *types.Project {
+				t.Helper()
+
+				secretsDir := filepath.Join(tmpDir, "secrets")
+				if err := os.Mkdir(secretsDir, filesystem.PermDir); err != nil {
+					t.Fatalf("failed to create %s: %v", secretsDir, err)
+				}
+
+				envFile := filepath.Join(secretsDir, "app.env")
+				copyFile(t, encryptedEnvSrc, envFile)
+
+				staticDir := filepath.Join(tmpDir, "static")
+				if err := os.Mkdir(staticDir, filesystem.PermDir); err != nil {
+					t.Fatalf("failed to create %s: %v", staticDir, err)
+				}
+
+				// #nosec G703 -- path is constructed from t.TempDir(), not user input
+				if err := os.WriteFile(filepath.Join(staticDir, "index.html"), []byte("<html></html>"), filesystem.PermOwner); err != nil {
+					t.Fatalf("failed to write plain file: %v", err)
+				}
+
+				return &types.Project{
+					WorkingDir: tmpDir,
+					Services: types.Services{
+						"svc1": {
+							Name:     "svc1",
+							EnvFiles: []types.EnvFile{{Path: envFile}},
+							// The directory with the encrypted file is walked first, the
+							// directory without encrypted files second.
+							Volumes: []types.ServiceVolumeConfig{
+								{Type: "bind", Source: secretsDir, Target: "/run/secrets"},
+								{Type: "bind", Source: staticDir, Target: "/srv"},
+							},
+						},
+					},
+				}
+			},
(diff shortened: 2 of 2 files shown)


kimdre (approved):
Thank you for the PR! :)

kimdre:
I released your fix in v0.117.1

PR #1841, opened 2026-09-10
feat(notification): filter commit changelog by files of the stack

dennislapchenko:
Implements what you described in discussion #1835: filter inside `GetCommitsBetween` with go-git `LogOptions.PathFilter`, built from resolved file set of the project, and no new config setting.

Problem: changelog took whole range between last deployed commit and new one, so in repo with many stacks every stack reported also commits of other stacks.

Approach: file collection is extracted out of `DecryptProjectFiles` into `serviceFiles`, so decryption and changelog work on same set and cannot drift apart. For the changelog that set is extended by compose files and build contexts (`resolvedProjectFiles`). `ProjectPathFilter` then converts those docker-host paths to repo-root-relative ones, because that is what go-git passes to the filter — files match exactly, directories (bind mount source, build context) match by prefix. `GetCommitsBetween` got a `pathFilter` parameter, and the cap of 50 counts commits after filtering.

One place worth your eyes: range is still taken from the unfiltered log first and the filtered log is limited to it. go-git filters by dropping non-matching commits from the iterator, boundary commit included, so single filtered walk loses where to stop and runs back through whole history. Boundary logic itself is untouched. `TestGetCommitsBetween_PathFilter` covers exactly that — with naive single walk it returns an ancient commit of the stack instead of one.

Empty file set means no filter, so changelog stays unfiltered as today. Same when the stack covers repo root (then every path is relevant anyway). Empty changelog would be worse than a noisy one, imo. Failure of filter build only logs a warning, changelog is best-effort as before.

Tested: `go build ./...` and `golangci-lint run ./...` clean. New unit tests in `internal/git` (filtered vs unfiltered range, boundary commit of another stack, single-file filter, cap applied after filtering) and `internal/docker` (exact file match, directory prefix, build context, paths outside repo, empty set, repo root, nil project). 49 test packages pass; only `TestDeployCompose`, `TestVerifySocketConnection` and `TestVerifyDockerHostConnectionContextCancellation/Unix` fail, they stat hardcoded `/var/run/docker.sock` and fail the same on clean `main`. Docs updated in `wiki/docs/Advanced/Notifications.md`.

### DIFF
--- a/internal/docker/project_files.go
+++ b/internal/docker/project_files.go
@@ -1,0 +1,261 @@
+package docker
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"github.com/compose-spec/compose-go/v2/types"
+
+	"github.com/kimdre/doco-cd/internal/common/types/set"
+	"github.com/kimdre/doco-cd/internal/filesystem"
+)
+
+// projectFile is a single path of the file set a compose project resolves to.
+// Paths are absolute and on the docker host, like the paths in types.Project.
+type projectFile struct {
+	Path string
+	// IsDir marks a directory whose whole content belongs to the project,
+	// e.g. a bind mounted directory or a build context.
+	IsDir bool
+}
+
+// projectFileSet collects projectFiles without duplicates.
+type projectFileSet struct {
+	index map[string]int
+	files []projectFile
+}
+
+func newProjectFileSet() *projectFileSet {
+	return &projectFileSet{index: make(map[string]int)}
+}
+
+// absProjectPath resolves path against the working directory of the project if it is relative.
+func absProjectPath(p *types.Project, path string) string {
+	if !filepath.IsAbs(path) {
+		path = filepath.Join(p.WorkingDir, path)
+	}
+
+	return filepath.Clean(path)
+}
+
+// add resolves path against the working directory of the project if it is relative
+// and adds it to the set.
+func (s *projectFileSet) add(p *types.Project, path string, isDir bool) {
+	if path == "" {
+		return
+	}
+
+	path = absProjectPath(p, path)
+
+	if i, ok := s.index[path]; ok {
+		// A path used as file and as directory is a directory.
+		s.files[i].IsDir = s.files[i].IsDir || isDir
+
+		return
+	}
+
+	s.index[path] = len(s.files)
+	s.files = append(s.files, projectFile{Path: path, IsDir: isDir})
+}
+
+// list returns the collected files sorted by path.
+func (s *projectFileSet) list() []projectFile {
+	slices.SortFunc(s.files, func(a, b projectFile) int {
+		return strings.Compare(a.Path, b.Path)
+	})
+
+	return s.files
+}
+
+// serviceFiles resolves the files the services of a compose project read at deploy time:
+// configs, secrets, bind mount sources, env files, dockerfiles and build secrets.
+// Bind mount sources are the only entries that can be a directory.
+//
+// DecryptProjectFiles and resolvedProjectFiles both build on this, so the file set they
+// work on cannot drift apart.
+func serviceFiles(p *types.Project) ([]projectFile, error) {
+	files := newProjectFileSet()
+
+	for _, s := range p.Services {
+		for _, cfg := range s.Configs {
+			if cfg.Source != "" {
+				if cfgConfig, ok := p.Configs[cfg.Source]; ok && cfgConfig.File != "" {
+					files.add(p, cfgConfig.File, false)
+				}
+			}
+		}
+
+		for _, secret := range s.Secrets {
+			if secret.Source != "" {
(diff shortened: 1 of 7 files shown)


internal/git/commits.go
```diff
@@ -113,41 +113,113 @@ func commitBoundary(repo *git.Repository, oldHash, newHash plumbing.Hash) map[pl
...
-			Hash:      c.Hash.String(),
-			ShortHash: c.Hash.String()[:DefaultShortSHALength],
-			Subject:   strings.TrimSpace(subject),
-			Author:    c.Author.Name,
-		})
+		inRange[c.Hash] = struct{}{}
 
-		return nil
+		return true
+	})
```
kimdre, line 201: In `GetCommitsBetween`, when a `pathFilter` is supplied, the first walk that builds `inRange` (from `newHash` to the boundary) has no cap.

Before this PR, the single walk stopped at either the boundary or `maxCommits`, whichever came first. Now the cap only applies to the second (filtered) walk.

The tricky part: this hits exactly the scenario the PR is meant to fix — a monorepo with several stacks, where one stack's `DeployedCommit` is far behind `HEAD` because only other stacks changed in between. In that case, the first walk has to read through the entire intervening history uncapped (and build a map of every hash in it) before any filtering/capping kicks in. That eats into the performance win of filtering and could get slow/memory-heavy on long histories.

Suggestion: put a safety ceiling on the first walk too (something well above 50, e.g. a few thousand), with a warning log if it's hit, instead of leaving it fully unbounded.

internal/docker/project_files.go
```diff
@@ -0,0 +1,261 @@
...
+			}
+
+			// Additional contexts also take non-path values like docker-image://alpine,
+			// which simply never match a path in the repository.
+			files.add(p, ctx, filesystem.IsDir(absProjectPath(p, ctx)))
+		}
+	}
+
+	return files.list(), nil
+}
```
kimdre, line 189: Minor issue in `resolvedProjectFiles` / `ProjectPathFilter`: additional build contexts that don't exist locally yet are treated as files rather than directories (via `filesystem.IsDir`), so they'd need an exact-path match instead of prefix match if files later show up under them. Edge case, probably fine.

kimdre (changes requested):
Thank you for your PR! I found two things that need to be addressed.

internal/git/commits.go
```diff
@@ -113,41 +113,113 @@ func commitBoundary(repo *git.Repository, oldHash, newHash plumbing.Hash) map[pl
...
-			subject = subject[:i]
+		return commits, nil
+	}
+
+	// go-git filters paths by dropping the non-matching commits from the log iterator,
+	// the boundary commit included. The walk would then miss where to stop and run
+	// through the whole history, so the range is taken from the unfiltered log first and
+	// the filtered log is limited to it. Walking the range twice only reads commit
+	// objects, the tree diffs of the filtered walk dominate either way.
+	inRange := make(map[plumbing.Hash]struct{})
```
kimdre, line 191: `inRange`'s type can probably be changed to a `Set`, see [internal/common/types/set/set.go](https://github.com/kimdre/doco-cd/blob/main/internal/common/types/set/set.go)

kimdre (changes requested):

internal/docker/changelog_layout_test.go
```diff
@@ -0,0 +1,666 @@
...
+		"GIT_CONFIG_GLOBAL=/dev/null", "GIT_CONFIG_SYSTEM=/dev/null",
+	)
+
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		t.Fatalf("git %s: %v\n%s", strings.Join(args, " "), err, out)
+	}
+
+	return string(out)
+}
```
kimdre: Please use go-git to create and work in git repos in your tests; If possible also create them in-memory. You can find usage examples in my other tests, for example here: https://github.com/kimdre/doco-cd/blob/main/internal/git/commits_internal_test.go#L40-L49

internal/git/commits.go
```diff
@@ -131,95 +140,112 @@ func newCommitInfo(c *object.Commit) CommitInfo {
...
+// boundedCommitIter ends a commit walk at the first boundary commit, and after
+// maxScannedCommits commits at the latest.
+//
+// The boundary commit is still passed on, because object.NewCommitPathIterFromIter diffs
+// every commit against the next commit of the source iterator. Without it the oldest
+// commit of the range would be diffed against an empty tree and always look like a change.
+// Callers drop it by the same boundary check.
+type boundedCommitIter struct {
+	src      object.CommitIter
+	boundary map[plumbing.Hash]struct{}
```
kimdre: Please use the [Set](https://github.com/kimdre/doco-cd/blob/main/internal/common/types/set/set.go) type for `boundary`.

kimdre (changes requested):
Thank you, I found two more things. Please also resolve the merge conflict. :)

kimdre:
Please also fix the merge conflict. Everything else looks good to me now.

dennislapchenko:
woops, done!

kimdre (approved):

PR #1861, opened 2026-09-15
fix(scheduler): do not read own stop window as drift in pre-deploy

### LINKED ISSUE #1856: Scheduled job `stop_services` makes the next poll redeploy the stack it stopped
Related to #1852 (same feature), but different problem.

Setup — `backup` stack, job service with:

```
cd.doco.job.enabled: "true"
cd.doco.job.schedule: "0 2 * * *"
cd.doco.job.execution_mode: "restart"
cd.doco.job.stop_services: "db/db"
```

Job is cold file copy of database directory, so `db` must be down while it runs.

Seen on v0.114.0, code path is unchanged on v0.118.0:

```
02:00:00.019  stopping services before scheduled job
02:00:11.268  restarting services after scheduled job
02:00:12.967  deploying stack   stack=db  recreate={"forced_services":[],"mode":"diverged"}
```

Nothing is actually recreated, container id and `Created` stay same. But full deploy cycle runs and "Deployment completed" notification is sent, every night, per box.

### What I think happens

- poll tick lands inside stop window, while doco-cd itself keeps `db` stopped
- `getDeployStatus` counts replica only for containers in `running` state (`internal/docker/service_utils.go:238`), so stopped service reports 0
- `db` has `restart: unless-stopped`, so `allowStoppedForRestartPolicy` is false and `CheckServiceMismatch` gives `replicas mismatch`, want 1 got 0
- `shouldSkipDeployment` bails on any mismatch (`internal/stages/stage_2_pre-deploy.go:381`), so deploy goes ahead
- by the time deploy reaches docker the service is back up, nothing left to recreate

Two things I got wrong at first, in case somebody reads logs same way:

- `mode: "diverged"` is not a verdict. `api.RecreateDiverged` is just the default recreate mode in `deployComposeRuntime`. Noise, not signal.
- I assumed factor is how long service needs to become ready. Looks wrong — mismatch is on docker `running` state, not on health. Factor is length of stop window against poll interval. One box with small database (copy is quick) almost never gets hit, two production boxes with bigger copy get it every night. On slowest one that deploy took 1m6s and ran while database was still starting up.

Machinery for this already exists, but only on other path: `schedulerHolds` + `schedulerStopHoldGracePeriod` (10s) in `internal/reconciliation/manager.go` suppress exactly this false positive for the docker event listener (`internal/reconciliation/reconciliation.go:444`). Poll path in stage 2 never asks it.

### Ask

doco-cd should not read restart it did itself as external drift. Options, whichever one code supports better:

1. after `startServicesForJob`, refresh deploy record of affected stack, so next poll compares against current reality
2. let stage 2 consult `schedulerHolds` for stacks touched by `stop_services`, same way event listener does
3. wait until restarted services are `running` before scheduler releases them, then no mid-restart window exists

wdyt, which direction you prefer? Can send PR.

Compose mode, not swarm, polling enabled.

dennislapchenko:
Fixes #1856. Option 2, as you picked.

`schedulerHolds` was consulted only by docker event listener. Pre-deploy stage now asks it too, so mismatch of service which doco-cd itself holds stopped is dropped before `shouldSkipDeployment`. Rest of mismatches stay, real drift deploys same as before.

Scope is same as event listener: compose mode only, since holds are not registered for swarm jobs. Post-release grace period counts as held, so poll which starts while service is still coming back is covered too.

Wiring is `Manager.IsSchedulerStopHeld`, and `stages` takes it as optional `SchedulerStopHolds` dependency - nil means check is off, so `stages` keeps no dependency on `reconciliation`.

Tests: `TestDropSchedulerHeldMismatches_SkipsDeploymentDuringJobStopWindow` reproduces the nightly case - stopped service with `restart: unless-stopped`, asserts unfiltered mismatch forces deploy and held one skips it. Plus context scoping, swarm, no-tracker, and grace period of `IsSchedulerStopHeld`.

Also small note in Job-Scheduling docs that stop window is not drift.

`go test ./...` is green except failures which are already there on clean `main` on my machine, all from hardcoded `/var/run/docker.sock` (mac, no such socket): `cmd/doco-cd`, `internal/api`, and `TestDeployCompose` / `TestVerifySocketConnection` / `TestVerifyDockerHostConnectionContextCancellation` in `internal/docker`. Same set before and after.

btw lmk if you prefer holds to be queried by labels instead of project/service names, was easy either way.


kimdre (approved):

kimdre:
Thanks for the PR :)

PR #1864, opened 2026-09-15
feat(external-secrets): interpolate external secrets' external secrets

tYiZQMr93eYPp:
Allows resolving external secrets that reference external secrets.
E.G. with OpenBao provider:
* `ANOTHER_TEST_SECRET: kv:store:name:another-test-secret` -> `123`
* `TEST_SECRET: kv:store:name:test-secret` -> `${ANOTHER_TEST_SECRET}` (Doesn't interpolate to `123`).

With this change, it would interpolate `TEST_SECRET: kv:store:name:test-secret` properly to `${ANOTHER_TEST_SECRET}`, and, then, to `123`.

---

EDIT: Needs further testing


internal/secretprovider/types/secret_ref.go
```diff
@@ -178,6 +178,99 @@ func InterpolateExternalSecretRefs(in map[string]ExternalSecretRef, enabled bool
 	return out, nil
 }
 
+// errUnknownResolvedSecretRef is an internal sentinel: the referenced name is not one of the
+// resolved external secrets, so the placeholder is left untouched rather than substituted.
+var errUnknownResolvedSecretRef = errors.New("not an external secret")
+
+// resolvedSecretRefPattern matches ${NAME} or $NAME, used to find references to other resolved
+// external secrets inside an already-resolved secret value.
+var resolvedSecretRefPattern = regexp.MustCompile(`\$\{([A-Za-z_][A-Za-z0-9_]*)\}|\$([A-Za-z_][A-Za-z0-9_]*)`)
```
tYiZQMr93eYPp: Need to be able to escape the `$` in case users want to use the literal dollar sign.
It's probably best to do it by duplicating the dollar sign: `$$`

internal/secretprovider/types/secret_ref_test.go
```diff
@@ -248,3 +248,103 @@ func TestInterpolateExternalSecretRefs(t *testing.T) {
...
+		want      map[string]string
+		wantError bool
+	}{
+		{
+			name: "chained reference resolved",
+			input: map[string]string{
+				"DB_PASSWORD": "hunter2",
+				"DB_HOST":     "db.internal",
+				"DB_URL":      "postgres://user:${DB_PASSWORD}@${DB_HOST}/mydb",
+			},
```
kimdre, line 266: Please fix the linter error.

wiki/docs/External-Secrets/index.md
kimdre, line 1: Please fix the merge conflict on this file.

kimdre (changes requested):
~There is one real edge case in your code:~

~If the Infisical SDK fails to resolve a reference, it may leave the raw `${NAME}` text in the value instead of erroring. Both features use identical `${NAME}` syntax. See the doco-cd wiki at _Validate reference permissions_: https://doco.cd/latest/External-Secrets/Infisical/#infisical-secret-references~

~If a user enables `INTERPOLATE_RESOLVED_SECRETS` and also happens to have a doco-cd `external_secrets:` entry named e.g. `DB_HOST`, an unresolved Infisical placeholder would get silently substituted with doco-cd's own `DB_HOST` secret (potentially a different project/env/path than Infisical intended), masking a real Infisical failure instead of surfacing it. See also their referencing documentation: https://infisical.com/docs/documentation/platform/secret-reference~

~Maybe we should to add a fast-fail check to the Infisical provider when the server-side interpolation did not work and returned the raw `${NAME}` text to handle this error case? However we need to keep in mind that some users might have secrets that legitimately contain literal `${SOMENAME}`-looking text (unrelated to actual references).~

~What do you think?~

Edit: I created a PR to fix this edge case on the main branch: https://github.com/kimdre/doco-cd/pull/1868

internal/secretprovider/types/secret_ref_test.go
tYiZQMr93eYPp: Fixed both, @kimdre.
Thanks for pointing them out!

kimdre (approved):

kimdre:
Thank you for the PR :)

PR #1869, opened 2026-09-16
feat(notification): attribute the deploy config to its stack in the changelog

dennislapchenko:
## Problem

The changelog path filter is built from the compose project only:

```go
pathFilter, filterErr := docker.ProjectPathFilter(s.Repository.PathExternal, s.Docker.Project)
```

`.doco-cd.<target>.yml` is not a compose file, so `types.Project` never names it and it is never in the filter's path set. A commit that touches only the deploy config therefore matches nothing and is dropped — and the notification arrives with an empty changelog, even though that file is what declares the stack, holds its image tags, and caused the deploy.

The layout suite already documented this, in `changelog_layout_test.go`:

> `note: "GAP: the deploy config is not part of the compose project, so its own change is not reported"`

This PR closes that one. The `include:` / `extends: file:` gaps are untouched — they need compose-go's include listener and are a separate change.

## Why it matters in practice

This is the commonest deploy shape in a GitOps repo. A tag bump — opened by Renovate or written by an app repo's CI — usually touches *only* the deploy config. Measured across the two deployment repos I maintain, **~59% of commits touch nothing else**, and for all of them the commit subject was the only human-readable description of what shipped.

There is also an asymmetry inside doco-cd today: the deploy *decision* hashes the fully resolved project and does see the pinned include's content and the interpolated tag, while the *changelog* uses repo paths of that same project and does not.

## Change

`ProjectPathFilter` takes extra repo paths alongside the project, and the post-deploy stage passes `Config.Internal.File`, which is already populated. Paths outside the repository are ignored exactly as project files are. The closure body is unchanged.

## Scope and trade-off

Bounded: a stack only notifies when it actually deployed, so a deploy-config commit can only be listed by a stack that a change in that file caused to deploy.

It can over-report across documents in one file — a commit changing only stack B's tags would be listed by stack A if A deployed in the same range for an unrelated reason. That seemed a better trade than diff-attributing individual YAML documents.

## Tests

- The GAP layout now asserts the commit **is** reported.
- New layout covers the other half: another stack's deploy config is still not reported.
- The `git log` oracle receives the same path set, otherwise the two disagree by construction.
- Small refactor: the repo-relative path conversion in `projectRepoPaths` became a helper, since the extra paths need identical handling.


kimdre (approved):

kimdre:
Thanks for the PR. :)

PR #1876, opened 2026-09-17
feat(self-update): let a single instance re-own itself via GitOps

### LINKED ISSUE #827: Manage doco-cd configuration changes/image updates using doco-cd itself

dennislapchenko:
Closes #827.

## What it does

doco-cd can now deploy the stack that contains its own container. One instance, no updater sidecar. Push that bumps doco-cd image is a deploy like any other.

Off by default, `SELF_UPDATE_ENABLED=true` to turn on. Without it, a deploy that would replace own container is now refused with an error — instead of what happens today, which is killing the process mid-deploy and leaving the stack broken.

Full user-facing writeup (setup, how update runs, caveats): [`examples/self-updating/README.md`](https://github.com/kimdre/doco-cd/blob/main/examples/self-updating/README.md). Wiki page rewritten too, two-instance setup kept there as alternative.

## Why it did not work before

Compose runs in-process here, `compose/v5` as library and not shell-out. So `service.Create` stops own container mid-plan and the remaining ops — remove, rename, start — never run. Result is old container exited and new one stuck in `created`. Restart policy does not save you, docker ignores it after an API stop.

## Decisions

Two strategies, `auto` picks:

- `scale_out` — creates second container from new config, health-gates it while still serving, then hands over. No downtime. Rollback is one remove, because nothing was stopped yet. Needs no `container_name` and no published ports, two containers cannot share either.
- `applier` — clones own container into throwaway `doco-cd apply-self`, which does the recreate from outside and restores previous container from snapshot if new one is unhealthy. Few seconds downtime, but works with `container_name`.

`doco-cd apply-self --bootstrap` creates the first container through the normal deploy path, so it already carries correct compose + `cd.doco.*` labels. Replaces the temp-instance dance people do today.

Handover is journaled on the data volume, so a crash at any point converges on next boot. Failed update is poisoned for that commit and not retried until a new commit lands, otherwise the retry logic loops on a broken version forever.

Design was checked against a real daemon before writing feature code (`internal/docker/self_update_integration_test.go`) — mainly that `RecreateNever` + scale 2 really leaves the diverged container running and creates a correctly labelled #2. If that was false, whole `scale_out` idea is dead.

On your concern in #827 about auto-updates hiding breaking changes: opt-in by default, docs push digest pinning + Renovate so a human still approves each release, and a bad version is rolled back and then not retried.

## Two bugs found by the new e2e

- drain used `Runs.CloseAndWait`, which cancels in-flight runs. A commit touching doco-cd and another stack could leave that other stack destroyed mid-recreate. Added `Runs.Drain` — closes admission, waits, does not cancel.
- `.gitignore` had bare `doco-cd` for the built binary, which also ignored the whole `cmd/doco-cd/` directory. Anchored it.

## On size

Honestly it is a lot of code for what it does. Journal, two strategies, poison list, bootstrap subcommand, 8 e2e scenarios. Most of it is not the update itself but the "what if it dies halfway" part, which for a thing that replaces the controller feels non-optional imo.

Could be trimmed to `applier` only and drop `scale_out`, maybe 40% less. But no-downtime path is the nicer one and it is where design is actually cleanest, so I kept both. lmk if you prefer only one.

Either way — defo convenient thing to have. Controller under same GitOps as everything else it deploys, nothing to babysit.

## Testing

- e2e: 8 new scenarios (bootstrap, scale-out, applier, rollback x2, poison, crash recovery x2, regression). Full suite green 4 runs in a row.
- unit: journal state machine, strategy selection, poison, applier clone, rollback, handover semantics.
- `golangci-lint run ./...` and `--build-tags e2e ./test/...` both clean.


kimdre:
I refactored how repositories are stored and handled inside doco-cd in v0.120.0. Could you please fix the merge conflicts before I begin my review?

kimdre (changes requested):
I refactored how repositories are stored and handled inside doco-cd in v0.120.0. Could you please fix the merge conflicts before I begin my review?

dennislapchenko:
done

kimdre:
I’ve pushed follow-up commits addressing the self-update review findings. They make journal transitions safe across processes, require applier readiness and completion of in-flight work before takeover, and improve cleanup and rollback after failed handovers or network recreation. They also validate the running container’s restart policy and reject existing named-volume changes that rollback cannot restore.

I updated the self-update documentation and example bootstrap to `0.121.0`. The wiki bootstrap command remains on `latest` as requested.

Please review these changes, particularly the handover and recovery behavior, network-drift rollback, and volume restriction.

PR #1879, opened 2026-09-18
docs(examples): env-first layout, shared/, and the CI bump job

dennislapchenko:
Reworked the `deployments-repo` example to match the layout we ended up with running this on a real fleet.

- **Env-first.** Was `shop/{shop-dev,env,secrets}/`, now `shop-dev/{app,db}/` where each stack holds its own compose, `env/` and `secrets/` and refers to them by plain relative paths. One directory is everything a host runs, and copying a stack dir is how you start a new one.
- **`shared/`** for composes that several envs include, pulled in at a pinned sha from the same repo. Covers stacks with no app repo of their own — `db` here, in practice also telemetry and backup.
- **`app-repo/.github/workflows/deploy.yml`** — build the image, then move the image tag and the compose sha together so the two always match. That loop was the part the example left to the reader.

README rewritten for someone meeting this for the first time: layout, how the pieces fit, five setup steps, and the four tasks people actually do (deploy code, change a shared compose, bump a third-party image, check a secret landed).


dennislapchenko:
@kimdre I invite you to add examples of your usages please :) 
I have a hunch you're controlling many doco-cd instances centrally, via API, or even single doco-cd for multiple hosts. I'd love to see how that setup works.
Also examples are missing anything swarmey, something i also never dealt with so would be cool to see.

PR #1880, opened 2026-09-18
fix(external-secrets): always load dotenv & ext. secrets

tYiZQMr93eYPp:
`s.DeployConfig.RepositoryUrl` is only being set when the when the deploy config points to a separate remote repo from the one that's triggering the sync.
This meant that dotenv and external secrets files wouldn't load properly when provisioning `doco-cd` from a `doco-cd` initializer.

It's also been optimized a little bit to prevent duplicating code and having an unnecessary `else` clause.

Apologies in advance.


kimdre:
Thanks! I added some tests to verify the outcome. :)

